PR #8, opened 2021-06-03
Add some 64-bit Burn tests.

barnson:


src/test/burn/WixToolsetTest.BurnE2E/BasicFunctionalityTests.cs
```diff
@@ -41,6 +41,12 @@ public void CanInstallAndUninstallSimpleBundle_x64_wixstdba()
             this.CanInstallAndUninstallSimpleBundle("PackageA_x64", "BundleA_x64");
         }
 
+        [Fact]
+        public void CanInstallAndUninstallSimplePerUserBundle_x64_wixstdba()
```
rseanhall, line 45: We need to get these tests running in an automated build. This test doesn't pass. The problem is here:

https://github.com/wixtoolset/wix4/blob/f0105569c28694d81e8e27a4c3a4cc1648676a06/src/test/burn/WixToolsetTest.BurnE2E/BasicFunctionalityTests.cs#L74

`package.GetInstalledFilePath` assumes the install folder is rooted in `ProgramFiles6432Folder` since that's what the per-machine template uses, but the per-user template uses `LocalAppDataFolder`.

barnson: Don't the tests run during the PR build?

rseanhall: No. That would add almost 10 minutes last time I checked to the build, and would just grow as we write more tests.

barnson: It already takes 8.5 hours, what's 10 minutes more? OK, well, I can't even build WiX atm, so I'll comment out the tests for now.

rseanhall: We probably should discuss this during the meeting if a core WiX developer can't build.

barnson: Well, I really want to get my traversal projects change in and keep adding to them so the build doesn't take so long, at least locally. But I haven't had time to bug @robmen yet.

PR #12, opened 2021-06-06
Issue4770

cpuwzd:
Extended the notion of falsehood in preprocessor expressions.

Added code to allow "", "0", "no", "false" and "off" (ignoring case, with or without quotes) to serve as alternative representations of falsehood in the preprocessor. In the case of zero, any number of digits are allowed

Fixes.wixtoolset/issues#4770.


barnson:
This is a breaking change. If `wix convert` can't fix authoring like this (and I suspect it can't), it at least needs to warn. I'll reopen https://github.com/wixtoolset/issues/issues/4770 for discussion at next week's meeting. (FTR, I'm not a fan of the preprocessor working this way.)

PR #14, opened 2021-06-09
5300 - Earlier Burn splash screen and dictionaries for payloads.

rseanhall:
Make the initial Burn process show the splash screen, and show it before parsing the manifest.

Fix building and perf for bundles with 1000s of package and BA payloads.

When building a bundle with two `PayloadGroup`s that each have 10000 payloads, one referenced in a package and the other for the BA, wix.exe crashed with an OutOfMemory exception due to trying to order all of those payloads. Then it crashed with a StackOverflow exception due to trying to serialize 10000 nested `PreviousValue`s.

The time from launching the bundle to DetectBegin using wixstdba was originally 10-11 seconds with the splash screen shown for less than a second. Now the splash screen is shown in less than a second and the time to DetectBegin is 3-4 seconds. Extracting the attached container during Apply went down from 7 seconds to 5 seconds.

Fixes https://github.com/wixtoolset/issues/issues/5300


src/burn/engine/payload.cpp
```diff
@@ -190,12 +190,46 @@ extern "C" HRESULT PayloadsParseFromXml(
...
+            BURN_PAYLOAD* pPayload = &pPayloads->rgPayloads[i];
+            BURN_CONTAINER* pContainer = pPayload->pContainer;
+
+            if (!pContainer)
+            {
+                continue;
+            }
+            else if (!pContainer->payloads.sdhPayloads)
+            {
+                //hr = MemAllocArray(reinterpret_cast<LPVOID*>(&pContainer->payloads.rgpPayloads), sizeof(BURN_PAYLOAD*), pContainer->cParsedPayloads);
```
barnson, line 215: Remove before commit?

src/burn/engine/payload.cpp
```diff
@@ -190,12 +190,46 @@ extern "C" HRESULT PayloadsParseFromXml(
...
+                //ExitOnFailure(hr, "Failed to allocate memory for container payloads.");
+
+                hr = DictCreateWithEmbeddedKey(&pContainer->payloads.sdhPayloads, pContainer->cParsedPayloads, NULL, offsetof(BURN_PAYLOAD, sczSourcePath), DICT_FLAG_NONE);
+                ExitOnFailure(hr, "Failed to create dictionary for container payloads.");
+            }
+
+            /*hr = MemEnsureArraySizeForNewItems(reinterpret_cast<LPVOID*>(&pContainer->payloads.rgpPayloads), pContainer->payloads.cPayloads, 1, sizeof(BURN_PAYLOAD*), 5);
```
barnson, line 222: And again?

PR #16, opened 2021-06-11
Allow access to persisted variables from related bundles #3704

jchoover:
Thoughts on the unit tests?

Also I have 2 options on the new butil call, one that allocates the string for the caller, and one that uses a pre-allocated buffer and length.

https://github.com/wixtoolset/issues/issues/3704


src/libs/dutil/WixToolset.DUtil/butil.cpp
```diff
@@ -29,10 +30,13 @@ NOTE: caller is responsible for closing key
...
-    __in BUNDLE_INSTALL_CONTEXT context, 
-    __inout HKEY *key);
-
+    __in BUNDLE_INSTALL_CONTEXT context,
+    __in_opt LPCWSTR szSubKey,
+    __inout HKEY* key);
 
+/********************************************************************
+BundleGetBundleInfo - Read the registration data for a gven bundle
+********************************************************************/
```
rseanhall, line 39: FYI, these documentation comments are in the .cpp file in many places but they're supposed to be in the .h file so the consumers get them.

rseanhall (approved):
This looks good. I'm not sure why GitHub didn't send a notification that it was ready to review. I'm going to fix some style nits before I merge this.

src/burn/engine/registration.cpp
```diff
@@ -1011,7 +1034,76 @@ extern "C" HRESULT RegistrationSaveState(
...
+        {
+            hr = S_OK;
+            break;
+        }
+
+        ExitOnFailure(hr, "Failed to enumerate value %u", i);
+
+        hr = RegDeleteValue(hkRegistration, sczValueName);
+        ExitOnFailure(hr, "Failed to delete registration variable value.");
+    }
```
rseanhall, line 1067: I think you originally added this because we were going to get rid of the state file and enumerate the values instead. Can we remove this code now? The bundle should be writing the exact same variables every time so the values should always be overwritten. We're not clearing the values for any of the other registration keys.

rseanhall:
I manually rebased and squashed this into https://github.com/wixtoolset/wix4/commit/f3c96bcab560cb09355e9366eac3f4195479d95d. Thanks!

PR #17, opened 2021-06-13
Use traversal projects to build libs in parallel.

barnson:


src/libs/libs.cmd
```diff
@@ -12,38 +12,7 @@
 
 nuget restore || exit /b
 
-:: dutil
-
-msbuild -t:Test -p:Configuration=%_C% dutil\test\DUtilUnitTest || exit /b
```
rseanhall, line 17: How do the tests get run now?

barnson: dutil.proj line 14

PR #25, opened 2021-07-05
Resolves wixtoolset/Issues/5371

cpuwzd:
Allows a user comment to be specified when creating a new user account during installation.

Creates a desktop folder containing test.msi and test.pdb. Installing test.msi creates user "Fred"
with password "Flintstone" and a pithy user comment. Uninstalling removes the account. Installation
fails if the account already exists, but I believe that that functionality is already optional. The failure
only occurs because of the options selected in the test.

The means of delivering the test.msi file might not be the most appropriate, and the whole operation raises
security questions. Ideally, testing could be automated in a virtual environment without exposing test.msi
and without issues of compromising the security or the sovereignty of the machine running the test.

It works for now, but perhaps the test should be disabled with appropriate comments until there is a reported
issue with the functionality of the fix itself.

Talk amongst yourselves!

### DIFF
--- a/src/ext/Util/wixext/UtilCompiler.cs
+++ b/src/ext/Util/wixext/UtilCompiler.cs
@@ -2680,18 +2680,18 @@ namespace WixToolset.Util
                             var bitnessValue = this.ParseHelper.GetAttributeValue(sourceLineNumbers, attrib);
                             switch (bitnessValue)
                             {
-                            case "always32":
-                                win64 = false;
-                                break;
-                            case "always64":
-                                win64 = true;
-                                break;
-                            case "default":
-                            case "":
-                                break;
-                            default:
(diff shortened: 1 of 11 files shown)


src/ext/Util/wixext/UtilCompiler.cs
```diff
@@ -3253,6 +3253,7 @@ private void ParseUserElement(Intermediate intermediate, IntermediateSection sec
             int attributes = 0;
             string domain = null;
             string name = null;
+            string comment = " Comment Not Set ";
```
barnson, line 3256: Should be empty.

barnson:
Custom action changes would also be needed...?

cpuwzd:
The intention here is to be able to distinguish the absence of a Comment 
attribute from the presence of a Comment attribute with an empty string 
as its value. The thought is that, if or when User is allowed when the 
user already exists,
it should be possible to remove a previously present comment. I have 
previously made it impossible to intentionally
create a comment with padding, so this value can be used as a sentinel. 
If I negelected to allow an empty Comment
Attribute value, that is my error.

cpuwzd:
It was not my intent to close the pull request before it was reviewed and merged.

PR #28, opened 2021-07-13
Prepares issue 5371 for review and comment.

cpuwzd:
Test produces a CreateUser folder on the desktop,
containing test.msi, which can be installed and
uninstalled for manual testing. This test is in need
of a virtual environment in which to safely run the
desired tests.

Tests should cover install and uninstall with
verification of user account (with comment) creation
and destruction.

Tests should also cover situations such as the user
account already existing and other options controlling
the possible options controlling the desired action.


cpuwzd:
Superseded by pull request #29.

PR #29, opened 2021-07-14
Prepares for review and discussion of Issue 5371

cpuwzd:
Test produces a CreateUser folder on the desktop,
containing test.msi, which can be installed and
uninstalled for manual testing. This test is in need
of a virtual environment in which to safely run the
desired tests.

Tests should cover install and uninstall with
verification of user account (with comment) creation
and destruction.

Tests should also cover situations such as the user
account already existing and other options controlling
the possible options controlling the desired action.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -19,7 +19,7 @@
     <OutputPath>$(BaseOutputPath)$(Configuration)\</OutputPath>
 
     <!-- This is C++-specific but used from managed projects to pick up native build artifacts. -->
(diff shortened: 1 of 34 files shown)


barnson (commented):
I'm specifically looking for another test like `CanBuildUsingFileShare` that queries the `Wix4User` table.

robmen:
@cpuwzd I would recommend splitting each of these commits into separate PRs. That was some (the smaller ones) can be accepted while others (bigger changes) are still under review.

PR #31, opened 2021-07-19
Fix more 32-bit assumptions

rseanhall:
Use strutil to return value in BundleGetBundleInfo.

### DIFF
--- a/src/libs/dutil/WixToolset.DUtil/butil.cpp
+++ b/src/libs/dutil/WixToolset.DUtil/butil.cpp
@@ -61,21 +61,16 @@ static HRESULT OpenBundleKey(
 DAPI_(HRESULT) BundleGetBundleInfo(
     __in_z LPCWSTR wzBundleId,
     __in_z LPCWSTR wzAttribute,
-    __out_ecount_opt(*pcchValueBuf) LPWSTR lpValueBuf,
-    __inout_opt LPDWORD pcchValueBuf
+    __deref_out_z LPWSTR* psczValue
     )
 {
-    Assert(wzBundleId && wzAttribute);
-
     HRESULT hr = S_OK;
-    LPWSTR sczValue = NULL;
     HKEY hkBundle = NULL;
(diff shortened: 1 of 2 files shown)


src/libs/dutil/WixToolset.DUtil/butil.cpp
```diff
@@ -147,7 +123,7 @@ DAPI_(HRESULT) BundleEnumRelatedBundle(
     HKEY hkUninstall = NULL;
     HKEY hkBundle = NULL;
     LPWSTR sczUninstallSubKey = NULL;
-    DWORD cchUninstallSubKey = 0;
+    size_t cchUninstallSubKey = 0;
```
barnson, line 126: We usually use `SIZE_T`, for whatever reason...

rseanhall: If it's a public parameter I used `SIZE_T` to follow our other APIs. Otherwise, I used `size_t` because the Windows APIs take `size_t` which is not implicitly convertible from `SIZE_T`.

src/libs/dutil/WixToolset.DUtil/butil.cpp
```diff
@@ -61,21 +61,16 @@ static HRESULT OpenBundleKey(
 DAPI_(HRESULT) BundleGetBundleInfo(
     __in_z LPCWSTR wzBundleId,
     __in_z LPCWSTR wzAttribute,
-    __out_ecount_opt(*pcchValueBuf) LPWSTR lpValueBuf,
-    __inout_opt LPDWORD pcchValueBuf
+    __deref_out_z LPWSTR* psczValue
```
barnson, line 64: Downside is that it requires strutil and C. Buffers are language-agnostic(er).

rseanhall: This is not exposed through the engine. This is a dutil API which means you're already using VS and C++.

barnson: Yeah but it complicates the translation layer if you want to use it outside of C. Not a big deal.

PR #35, opened 2021-08-23
On SuppressDowngradeFailure, quit after detect.

barnson:
Implements https://github.com/wixtoolset/issues/issues/6537.


rseanhall:
It's weird that you added a test bundle under `src/test/burn/TestData/WixStdBaTests` but didn't add a test that runs it.

PR #37, opened 2021-08-30
Add NativeMachine to Burn and WIX_NATIVE_MACHINE to UtilExtension

ericstj:
Add the ability for bundles and MSI's to detect when running on an ARM64 machine.

This directly plumbs the result of https://docs.microsoft.com/en-us/windows/win32/api/wow64apiset/nf-wow64apiset-iswow64process2 `pNativeMachine` argument, which is documented as 
> On success, returns a pointer to a possible [IMAGE_FILE_MACHINE_*](https://docs.microsoft.com/en-us/windows/win32/sysinfo/image-file-machine-constants) value identifying the native architecture of host system.

Issue: https://github.com/wixtoolset/issues/issues/6556
Wix3: https://github.com/wixtoolset/wix3/pull/538

### DIFF
--- a/src/burn/engine/variable.cpp
+++ b/src/burn/engine/variable.cpp
@@ -97,6 +97,10 @@ static HRESULT InitializeVariableVersionNT(
     __in DWORD_PTR dwpData,
     __inout BURN_VARIANT* pValue
     );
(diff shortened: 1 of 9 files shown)


barnson:
Thanks!

PR #38, opened 2021-09-14
Make heat use passed in extensions

InKahootz:
While the command line service was parsing and building a list of extensions to be passed to the heat tool, they weren't actually being used in the heat command line.

The extension manager now loads all passed-in extensions and gets any appropriate `IHeatExtension`s.  
This PR will also open up all built-in heat extensions, harvesters, and mutators for external use like they were in WiX3.

Fixes: wixtoolset/issues#6564

### DIFF
--- a/src/wix/heat/AssemblyHarvester.cs
+++ b/src/wix/heat/AssemblyHarvester.cs
@@ -10,7 +10,7 @@ namespace WixToolset.Harvesters
     /// <summary>
     /// Harvest WiX authoring from an assembly file.
     /// </summary>
-    public class AssemblyHarvester
+    public sealed class AssemblyHarvester
     {
         /// <summary>
         /// Harvest the registry values written by RegisterAssembly.
--- a/src/wix/heat/AssemblyHarvester.cs
+++ b/src/wix/heat/AssemblyHarvester.cs
@@ -10,7 +10,7 @@ namespace WixToolset.Harvesters
     /// <summary>
     /// Harvest WiX authoring from an assembly file.
     /// </summary>
-    internal class AssemblyHarvester
+    public class AssemblyHarvester
     {
         /// <summary>
         /// Harvest the registry values written by RegisterAssembly.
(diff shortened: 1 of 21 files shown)


InKahootz:
I wasn't positive if I should be packing heat.csproj. I'll receive feedback on this first.

Otherwise, I'm having to reference heat.dll in a roundabout way for my heat extension project. And that's just the netcore dll. The framework heat is an exe, which makes the situation worse.

```xml
  <ItemGroup>
    <Reference Include="heat.dll" Private="false" Publish="false" ExcludeAssets="runtime">
      <HintPath>$(PkgWixToolset_Sdk)\tools\netcoreapp3.1\%(Identity)</HintPath>
    </Reference>
  </ItemGroup>
```

rseanhall:
How are you using the newly public types? Can they be sealed?

The reference problem is really a question for @robmen. I don't know how he expected someone to create a netstandard2.0 dll for their extension when the types are only available from either a .NET Framework .exe or a netcoreapp .dll. I believe at runtime both .NET Framework and .NET Core will look for <assembly>.dll first and eventually fallback to <assembly>.exe.

The packing question is not specific to Heat, there's a similar distribution question for makesfxca.exe, thmviewer.exe, and possibly others (insignia?).

InKahootz:
Sealing them would be no problem for me. I just need to instantiate them for my extension and utilize their existing harvesting capability.

I'm curious though. Why? I've seen the PRs in [the dotnet/runtime repo](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#peanut-butter) where many internal/private classes are sealed by default for performance reasons.

rseanhall:
I unsealed them when I made them internal to match what we were doing elsewhere in https://github.com/wixtoolset/Harvesters/commit/b8029fe78c15790b1152e4a11fb867be6fed7576 (I guess Heat's history got destroyed when moving back to the monorepo). It seems we're not modernizing them into interfaces so they should be put back to how they were in v3. We can unseal them later if needed, but we can't seal them later because that's a breaking change.

rseanhall:
Thanks!

PR #40, opened 2021-09-30
Support multiple attached containers

nirbar:
See https://github.com/wixtoolset/issues/issues/6144

### DIFF
--- a/src/burn/engine/section.cpp
+++ b/src/burn/engine/section.cpp
@@ -8,7 +8,7 @@
 // If these defaults ever change, be sure to update constants in burn\stub\StubSection.cpp as well.
 #define BURN_SECTION_NAME ".wixburn"
 #define BURN_SECTION_MAGIC 0x00f14300
-#define BURN_SECTION_VERSION 0x00000002
+#define BURN_SECTION_VERSION 0x00000003
 #define MANIFEST_CABINET_TOKEN L"0"
 
 // structs
@@ -26,7 +26,7 @@ typedef struct _BURN_SECTION_HEADER
 
     DWORD dwFormat;
     DWORD cContainers;
-    DWORD rgcbContainers[1];
+    DWORD rgcbContainers[116];
 } BURN_SECTION_HEADER;
 
 static HRESULT VerifySectionMatchesMemoryPEHeader(
--- a/src/burn/stub/StubSection.cpp
+++ b/src/burn/stub/StubSection.cpp
@@ -7,7 +7,7 @@
 // If these defaults ever change, be sure to update constants in burn\engine\section.cpp as well.
 #pragma data_seg(push, ".wixburn")
 static DWORD dwMagic = 0x00f14300;
-static DWORD dwVersion = 0x00000002;
+static DWORD dwVersion = 0x00000003;
 
 static GUID guidBundleId = { };
 
@@ -18,6 +18,5 @@ static DWORD dwOriginalSignatureSize = 0;
 
 static DWORD dwContainerFormat = 1;
 static DWORD dwContainerCount = 0;
-static DWORD qwBootstrapperApplicationContainerSize = 0;
-static DWORD qwAttachedContainerSize = 0;
+static DWORD qwAttachedContainerSizes[116]; // Including UX container
 #pragma data_seg(pop)
--- a/src/wix/WixToolset.Core.Burn/Bind/BindBundleCommand.cs
+++ b/src/wix/WixToolset.Core.Burn/Bind/BindBundleCommand.cs
@@ -296,15 +296,13 @@ namespace WixToolset.Core.Burn
                 }
 
                 // Give the embedded payloads without an embedded id yet an embedded id.
-                var payloadIndex = 0;
                 foreach (var payload in payloadSymbols.Values)
                 {
                     Debug.Assert(PackagingType.Unknown != payload.Packaging);
 
                     if (PackagingType.Embedded == payload.Packaging && String.IsNullOrEmpty(payload.EmbeddedId))
                     {
-                        payload.EmbeddedId = String.Format(CultureInfo.InvariantCulture, BurnCommon.BurnAuthoredContainerEmbeddedIdFormat, payloadIndex);
-                        ++payloadIndex;
+                        payload.EmbeddedId = Guid.NewGuid().ToString("N");
                     }
                 }
             }
--- a/src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
+++ b/src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
@@ -3,6 +3,7 @@
 namespace WixToolset.Core.Burn.Bundles
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
     using WixToolset.Data;
@@ -19,7 +20,6 @@ namespace WixToolset.Core.Burn.Bundles
     {
         public const string BurnNamespace = "http://wixtoolset.org/schemas/v4/2008/Burn";
         public const string BurnUXContainerEmbeddedIdFormat = "u{0}";
-        public const string BurnAuthoredContainerEmbeddedIdFormat = "a{0}";
 
         public const string BADataFileName = "BootstrapperApplicationData.xml";
         public const string BADataNamespace = "http://wixtoolset.org/schemas/v4/BootstrapperApplicationData";
@@ -67,7 +67,7 @@ namespace WixToolset.Core.Burn.Bundles
         //   40-43:  container type (1 = CAB)
         //   44-47:  container count
         //   48-51:  byte count of manifest + UX container
-        //   52-55:  byte count of attached container
+        //   52-512:  byte count of attached containers (4 bytes for each container)
         protected const UInt32 BURN_SECTION_OFFSET_MAGIC = 0;
         protected const UInt32 BURN_SECTION_OFFSET_VERSION = 4;
         protected const UInt32 BURN_SECTION_OFFSET_BUNDLEGUID = 8;
@@ -78,11 +78,14 @@ namespace WixToolset.Core.Burn.Bundles
         protected const UInt32 BURN_SECTION_OFFSET_FORMAT = 40;
         protected const UInt32 BURN_SECTION_OFFSET_COUNT = 44;
         protected const UInt32 BURN_SECTION_OFFSET_UXSIZE = 48;
-        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE = 52;
-        protected const UInt32 BURN_SECTION_SIZE = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE + 4; // last field + sizeof(DWORD)
+        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0 = 52;
 
         protected const UInt32 BURN_SECTION_MAGIC = 0x00f14300;
-        protected const UInt32 BURN_SECTION_VERSION = 0x00000002;
+        protected const UInt32 BURN_SECTION_VERSION = 0x00000003;
+        protected const UInt32 BURN_SECTION_COMPATIBLE_VERSION = 0x00000002;
+        protected const UInt32 BURN_SECTION_SIZE = 512;
+        protected const UInt32 BURN_SECTION_MAX_ATTACHEDCONTAINER_COUNT = (BURN_SECTION_SIZE - BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0) / sizeof(UInt32);
+
         protected string fileExe;
         protected UInt32 peOffset = UInt32.MaxValue;
         protected UInt16 sections = UInt16.MaxValue;
@@ -112,6 +115,7 @@ namespace WixToolset.Core.Burn.Bundles
         {
             this.Messaging = messaging;
             this.fileExe = fileExe;
+            this.AttachedContainers = new List<ContainerSlot>();
         }
 
         public UInt32 Checksum { get; protected set; }
@@ -126,8 +130,7 @@ namespace WixToolset.Core.Burn.Bundles
         public UInt32 ContainerCount { get; protected set; }
         public UInt32 UXAddress { get; protected set; }
         public UInt32 UXSize { get; protected set; }
-        public UInt32 AttachedContainerAddress { get; protected set; }
-        public UInt32 AttachedContainerSize { get; protected set; }
+        public List<ContainerSlot> AttachedContainers { get; protected set; }
 
         protected IMessaging Messaging { get; }
 
@@ -177,7 +180,9 @@ namespace WixToolset.Core.Burn.Bundles
             }
 
             reader.BaseStream.Seek(this.wixburnDataOffset, SeekOrigin.Begin);
-            byte[] bytes = reader.ReadBytes((int)BURN_SECTION_SIZE);
+            List<byte> manifest = new List<byte>();
+            manifest.AddRange(reader.ReadBytes((int)BURN_SECTION_SIZE));
+            byte[] bytes = manifest.ToArray();
             UInt32 uint32 = 0;
 
             uint32 = BurnCommon.ReadUInt32(bytes, BURN_SECTION_OFFSET_MAGIC);
@@ -188,9 +193,9 @@ namespace WixToolset.Core.Burn.Bundles
             }
 
             this.Version = BurnCommon.ReadUInt32(bytes, BURN_SECTION_OFFSET_VERSION);
-            if (BURN_SECTION_VERSION != this.Version)
+            if ((BURN_SECTION_VERSION != this.Version) && (BURN_SECTION_COMPATIBLE_VERSION != this.Version))
             {
-                this.Messaging.Write(ErrorMessages.BundleTooNew(this.fileExe, this.Version));
+                this.Messaging.Write(BurnBackendErrors.IncompatibleWixBurnSection(this.fileExe, this.Version));
                 return false;
             }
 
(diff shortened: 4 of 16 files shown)


src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -78,8 +78,8 @@ internal abstract class BurnCommon : IDisposable
         protected const UInt32 BURN_SECTION_OFFSET_FORMAT = 40;
         protected const UInt32 BURN_SECTION_OFFSET_COUNT = 44;
         protected const UInt32 BURN_SECTION_OFFSET_UXSIZE = 48;
-        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE = 52;
-        protected const UInt32 BURN_SECTION_SIZE = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE + 4; // last field + sizeof(DWORD)
+        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0 = 52;
+        protected const UInt32 BURN_SECTION_MIN_SIZE = BURN_SECTION_OFFSET_UXSIZE + 4; // last field is the UX container, as containers could be authored external. + sizeof(DWORD)
```
rseanhall, line 82: I'm missing something here. We were predefining the `.wixburn` section in src\burn\stub\StubSection.cpp with all the necessary fields. You didn't touch that file. It was a constant size. We can just write as much data as we want into that section?

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -224,8 +231,21 @@ protected bool Initialize(BinaryReader reader)
...
-            this.AttachedContainerSize = this.ContainerCount > 1 ? BurnCommon.ReadUInt32(bytes, BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE) : 0;
+            this.AttachedContainers.Clear();
+            uint nextAddress = this.EngineSize;
+            if (this.ContainerCount > 1)
+            {
+                for (uint i = 0; i < (this.ContainerCount - 1 /* Excluding UX */); ++i)
+                {
+                    uint sizeOffset = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0 + (i * 4);
+                    uint size = BurnCommon.ReadUInt32(bytes, sizeOffset);
+                    if (size > 0)
```
rseanhall, line 242: Probably not a functional change, but I find these special cases for 0 byte containers confusing. If you're trying to optimize something here, then it should be in the binder not in this low level code.

src/wix/WixToolset.Core.Burn/Bundles/BurnWriter.cs
```diff
@@ -138,10 +143,19 @@ public bool AppendContainer(Stream containerStream, long containerSize, BurnComm
                     break;
 
                 case Container.Attached:
-                    burnSectionCount = 2;
-                    burnSectionOffsetSize = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE;
+                    burnSectionCount = 2 + (uint)this.AttachedContainers.Count;
+                    burnSectionOffsetSize = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0 + ((uint)this.AttachedContainers.Count * 4);
```
rseanhall, line 147: `burnSectionCount` is a bundle level value. I guess we've been setting it multiple times all along so this doesn't really matter, but it's unnecessary to do it for every container.

`burnSectionOffsetSize` is a container level value that is an offset to the location of its size. `this.AttachedContainers.Count` needs to be replaced with the container's index so each container's size is written to a different location.

src/wix/WixToolset.Core.TestPackage/BundleExtractor.cs
```diff
@@ -50,14 +50,12 @@ public static ExtractBAContainerResult ExtractBAContainer(IMessaging messaging,
...
-        /// <param name="tempFolderPath">Temp path for extraction.</param>
         /// <returns>True if there was an attached container.</returns>
-        public static bool ExtractAttachedContainer(IMessaging messaging, string bundleFilePath, string destinationFolderPath, string tempFolderPath)
+        public static bool ExtractAttachedContainers(IMessaging messaging, string bundleFilePath, string destinationFolderPath)
         {
-            Directory.CreateDirectory(tempFolderPath);
             using (var burnReader = BurnReader.Open(messaging, bundleFilePath))
             {
-                return burnReader.ExtractAttachedContainer(destinationFolderPath, tempFolderPath);
+                return burnReader.ExtractAttachedContainers(destinationFolderPath);
```
rseanhall, line 58: Each attached container needs to be extracted to its own directory to prevent collisions.

rseanhall (commented):
Thanks for working on this. Please add an integration test to src/test/burn/WixToolsetTest.BurnE2E. I expect it would currently fail at runtime due to the issue of not writing the container sizes individually to the `.wixburn` section.

Is it possible to create a bundle with so many attached containers that it can't run because the .exe is too big?

src/wix/WixToolset.Core.Burn/Bind/BindBundleCommand.cs
```diff
@@ -296,15 +296,13 @@ public void Execute()
...
-                var payloadIndex = 0;
                 foreach (var payload in payloadSymbols.Values)
                 {
                     Debug.Assert(PackagingType.Unknown != payload.Packaging);
 
                     if (PackagingType.Embedded == payload.Packaging && String.IsNullOrEmpty(payload.EmbeddedId))
                     {
-                        payload.EmbeddedId = String.Format(CultureInfo.InvariantCulture, BurnCommon.BurnAuthoredContainerEmbeddedIdFormat, payloadIndex);
-                        ++payloadIndex;
+                        payload.EmbeddedId = Guid.NewGuid().ToString("N");
```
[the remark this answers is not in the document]
nirbar: Attached containers are extracted to the same folder with file names taken from the embedded id, later renamed to their correct file names. 
When you have multiple attached containers that all have the same payload ids ('a0', 'a1', ...) these payloads overwrite each other.
The unique id ensures that doesn't happen

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
nirbar: We didn't move any of the fields in that section, so no need to change it.
The dwContainerCount field had a constant value of 2 (UX + WixAttachedContainer).
The nice thing about the fix is that burn was behaving correctly, only the linking process had to be fixed

src/wix/WixToolset.Core.Burn/Bind/BindBundleCommand.cs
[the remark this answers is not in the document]
rseanhall: But the ids were already unique. `payloadIndex` was used across all the containers, it wasn't reset to 0 for each container.

(It's true that dark.exe extracts the attached containers that way. But since it renames the extracted files to the correct names before moving on to the next container, non-unique ids wouldn't have been a problem anyway.)

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
rseanhall: > The dwContainerCount field had a constant value of 2 (UX + WixAttachedContainer).

Not really relevant to my comment, but if there isn't an attached container, it would be 1 not 2.

> We didn't move any of the fields in that section, so no need to change it.

You're not moving fields but you are adding more fields to the end. There used to be one DWORD at the end for the attached container's size, but now there is a DWORD field for each attached container's size. If this does actually work where the binder can extend the `.wixburn` section as much as it wants, then we should simplify src\burn\stub\StubSection.cpp so it doesn't look like it needs to stay in sync with the binder.

src/wix/WixToolset.Core.Burn/Bundles/BurnWriter.cs
```diff
@@ -138,10 +143,19 @@ public bool AppendContainer(Stream containerStream, long containerSize, BurnComm
                     break;
 
                 case Container.Attached:
-                    burnSectionCount = 2;
-                    burnSectionOffsetSize = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE;
+                    burnSectionCount = 2 + (uint)this.AttachedContainers.Count;
```
rseanhall, line 146: The hardcoded 2 was for UX container + attached container. This needs to be `1 + (uint)this.AttachedContainers.Count`.

src/wix/WixToolset.Core.Burn/Bundles/BurnWriter.cs
nirbar: AttachedContainers.Count is incremented on each call, so both values should be updated

src/wix/WixToolset.Core.Burn/Bind/BindBundleCommand.cs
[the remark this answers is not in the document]
nirbar: It was 0 for each container in wix3, I didn't check in v4.

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
rseanhall: If you look at the [PE file format](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format), there's more work we should be doing if our section is not a constant size. We have assumptions in our code that are no longer valid, like the fact that the section would always fit in the smallest possible space. The attached container information is already stored in Burn's manifest, can we stop storing the number of attached containers and their sizes in our section header and require parsing the manifest for that information? This current implementation is relying on the fact that the linker is padding the section with extra bytes, and that we are not checking whether we're reading past the end of our section.

nirbar: Perhaps the easier way would be to limit attached containers number so it doesn't overflow the section size.
Say 10 containers should be way over the top since a x86 executable is bound to be less than 4GB, so 2 containers should fit almost all scenarios

rseanhall: I wouldn't want to spend my time putting in an artificial limit, but if you think that's easier then that's fine. If you do go that route then I would prefer that the Burn runtime code uses the section size as the limit instead of hardcoding something.

As far as what that limit should be, here's some things to think about.
* We support x64 and ARM64 in v4. I don't think 4GB is the limit for those executables. Also, people might need more containers because of the number of files (max 65535 per cab) instead of the size.
* The minimum size for the section seems to be 512 bytes. If so, we could support 115 ((512 - 52)/4) attached containers without increasing the size of the executable.

src/burn/engine/section.cpp
```diff
@@ -26,7 +26,7 @@ typedef struct _BURN_SECTION_HEADER
 
     DWORD dwFormat;
     DWORD cContainers;
-    DWORD rgcbContainers[1];
+    DWORD rgcbContainers[116];
```
rseanhall, line 29: This can stay declared as `[1]`.

src/burn/stub/StubSection.cpp
```diff
@@ -18,6 +18,5 @@ static DWORD dwOriginalSignatureSize = 0;
 
 static DWORD dwContainerFormat = 1;
 static DWORD dwContainerCount = 0;
-static DWORD qwBootstrapperApplicationContainerSize = 0;
-static DWORD qwAttachedContainerSize = 0;
+static DWORD qwAttachedContainerSizes[116]; // Including UX container
```
rseanhall, line 21: Probably should have comment explaining the math to get to 512 bytes, the minimum size of a section.

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -19,7 +20,6 @@ internal abstract class BurnCommon : IDisposable
     {
         public const string BurnNamespace = "http://wixtoolset.org/schemas/v4/2008/Burn";
         public const string BurnUXContainerEmbeddedIdFormat = "u{0}";
-        public const string BurnAuthoredContainerEmbeddedIdFormat = "a{0}";
```
rseanhall, line 22: Need to add a test to verify that either:
* This is change is required to prevent collisions or
* This change is not required, so it can be reverted along with the test changes

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -78,11 +78,14 @@ internal abstract class BurnCommon : IDisposable
...
-        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE = 52;
-        protected const UInt32 BURN_SECTION_SIZE = BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE + 4; // last field + sizeof(DWORD)
+        protected const UInt32 BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0 = 52;
 
         protected const UInt32 BURN_SECTION_MAGIC = 0x00f14300;
-        protected const UInt32 BURN_SECTION_VERSION = 0x00000002;
+        protected const UInt32 BURN_SECTION_VERSION = 0x00000003;
+        protected const UInt32 BURN_SECTION_COMPATIBLE_VERSION = 0x00000002;
+        protected const UInt32 BURN_SECTION_SIZE = 512;
+        protected const UInt32 BURN_SECTION_MAX_ATTACHEDCONTAINER_COUNT = (BURN_SECTION_SIZE - BURN_SECTION_OFFSET_ATTACHEDCONTAINERSIZE0) / sizeof(UInt32);
```
rseanhall, line 87: Investigate using the actual section size (`SizeOfRawData`) instead of using hardcoded values.

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -126,8 +130,7 @@ public BurnCommon(IMessaging messaging, string fileExe)
         public UInt32 ContainerCount { get; protected set; }
         public UInt32 UXAddress { get; protected set; }
         public UInt32 UXSize { get; protected set; }
-        public UInt32 AttachedContainerAddress { get; protected set; }
-        public UInt32 AttachedContainerSize { get; protected set; }
+        public List<ContainerSlot> AttachedContainers { get; protected set; }
```
rseanhall, line 133: Investigate whether the code would be simplified if the UX container was in this as well.

src/wix/WixToolset.Core.Burn/Bundles/BurnCommon.cs
```diff
@@ -177,7 +180,9 @@ protected bool Initialize(BinaryReader reader)
             }
 
             reader.BaseStream.Seek(this.wixburnDataOffset, SeekOrigin.Begin);
-            byte[] bytes = reader.ReadBytes((int)BURN_SECTION_SIZE);
+            List<byte> manifest = new List<byte>();
+            manifest.AddRange(reader.ReadBytes((int)BURN_SECTION_SIZE));
+            byte[] bytes = manifest.ToArray();
```
rseanhall, line 185: There's no reason anymore to create the `List<byte>`.

src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
```diff
@@ -168,16 +170,19 @@ public bool ExtractAttachedContainer(string outputDirectory, string tempDirector
...
             Directory.CreateDirectory(outputDirectory);
-            string tempCabPath = Path.Combine(tempDirectory, "attached.cab");
-
-            this.binaryReader.BaseStream.Seek(this.AttachedContainerAddress, SeekOrigin.Begin);
-            using (Stream tempCab = File.Open(tempCabPath, FileMode.Create, FileAccess.Write))
+            foreach (ContainerSlot cntnr in this.AttachedContainers)
             {
-                BurnCommon.CopyStream(this.binaryReader.BaseStream, tempCab, (int)this.AttachedContainerSize);
-            }
+                string tempCabPath = Path.GetTempFileName();
```
rseanhall, line 175: Should bring back the directory parameter because the Core integration tests won't be able to delete this file.

src/wix/WixToolset.Core.Burn/Bundles/BurnWriter.cs
```diff
@@ -127,6 +130,11 @@ public bool AppendContainer(Stream containerStream, long containerSize, BurnComm
             UInt32 burnSectionCount = 0;
             UInt32 burnSectionOffsetSize = 0;
 
+            if (containerSize == 0)
```
rseanhall, line 133: Investigate all of these checks for 0 byte containers. It seems like they were only necessary to catch when there was no attached container, but it should have checked the number of containers for that.

src/wix/WixToolset.Core.Burn/Bundles/BurnWriter.cs
```diff
@@ -138,10 +146,19 @@ public bool AppendContainer(Stream containerStream, long containerSize, BurnComm
...
                     // TODO: verify that the size in the section data is 0 or the same size.
-                    this.AttachedContainerSize = (uint)containerSize;
+                    uint nextAddress = this.EngineSize;
+                    foreach (ContainerSlot cntnr in this.AttachedContainers)
+                    {
+                        if (cntnr.Address >= nextAddress)
+                        {
+                            nextAddress = cntnr.Address + cntnr.Size;
+                        }
+                    }
```
rseanhall, line 157: Recalculating `nextAddress` for every container seems unnecessary. Investigate turning `AttachedContainers` to `UInt32[]` and keeping `nextAddress` as a class variable.

src/wix/WixToolset.Core.Burn/Inscribe/InscribeBundleCommand.cs
```diff
@@ -29,17 +29,23 @@ public bool Execute()
...
                 FileSystem.CopyFile(this.Context.SignedEngineFile, tempFile, allowHardlink: false);
-
-                // If there was an attached container on the original (unsigned) bundle, put it back.
-                if (reader.AttachedContainerSize > 0)
+                using (BurnWriter writer = BurnWriter.Open(this.Messaging, tempFile))
                 {
-                    reader.Stream.Seek(reader.AttachedContainerAddress, SeekOrigin.Begin);
+                    if (reader.Version != writer.Version)
+                    {
+                        this.Messaging.Write(BurnBackendErrors.IncompatibleWixBurnSection(this.Context.InputFilePath, reader.Version));
```
rseanhall, line 36: This should probably have its own error message.

PR #47, opened 2021-11-02
Follow up for multiple attached container support

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/6144


rseanhall:
Unfortunately, builds are failing due to the VS bug on both Appveyor and Github Actions now. I built Debug and Release successfully on my machine.

PR #49, opened 2021-11-10
Put back protection for wixstdba controls from thmutil declarative behavior

rseanhall:
Move ImageList element back under Theme since it's not a control.


rseanhall:
The build passed on Appveyor: https://ci.appveyor.com/project/wixtoolset/wix4/builds/41486155

PR #50, opened 2021-11-17
3421 - Reinstall packages that were uninstalled during failed upgrade bundle

rseanhall:
When rolling back a bundle failure, reinstall all upgrade related bundles.

Fixes https://github.com/wixtoolset/issues/issues/3421
https://github.com/wixtoolset/issues/issues/5796


src/burn/engine/plan.cpp
```diff
@@ -1487,6 +1524,39 @@ extern "C" HRESULT PlanRelatedBundlesComplete(
             hr = DependencyPlanPackageComplete(&pRelatedBundle->package, pPlan);
             ExitOnFailure(hr, "Failed to complete plan dependency actions for related bundle package: %ls", pRelatedBundle->package.sczId);
         }
+
+        if (fInstallingAnyPackage && BOOTSTRAPPER_RELATION_UPGRADE == pRelatedBundle->relationType)
```
rseanhall, line 1528: 1. There is no concept of "the older bundle", one bundle could have multiple bundles to uninstall as part of the Update related bundle behavior.

rseanhall, line 1528: 2. As you can probably see from the first commit, I was planning on using the operation to only select MajorUpgrade bundles. But some bundles could support downgrading, so the bundle to install is actually newer and would have been labeled as Downgrade.

src/burn/engine/apply.cpp
```diff
@@ -739,6 +744,9 @@ extern "C" HRESULT ApplyExecute(
                 IgnoreRollbackError(hrRollback, "Failed rollback actions");
             }
 
+            hrRollback = DoRestoreRelatedBundleActions(pEngineState, &context, pRestart);
```
rseanhall, line 747: 3. Identifying which upgrade bundle was responsible for each MSI that was uninstalled is not always possible. Another wrinkle here is that Exe packages could also be breaking things that the older bundles needs to repair.

rseanhall (commented):
As we discussed in [meeting 217](https://www.firegiant.com/blog/2021/8/5/wix-online-meeting-217-highlights/), the plan was to automatically trigger a normal install operation for the older bundle on rollback of an upgrade. Unfortunately, things aren't that black and white for Burn. I had to make Burn install all upgrade related bundles at the end of rollback regardless of which packages were executed for the following reasons:

PR #51, opened 2021-11-29
Fixes required to get build working using MSBuild64 from VS2022

robmen:


barnson:
:shipit:

robmen:
Updated skip message based on feedback.

PR #53, opened 2021-12-02
Always populate bind variables for file version and language

robmen:
Always populating the values (using "" for null) makes it easier to
diagnose when you finally get the bind variable syntax correct.
Previously, incorrect syntax and not having a value both resulted in
errors.

### DIFF
--- a/src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
+++ b/src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
@@ -2,6 +2,7 @@
 
 namespace WixToolsetTest.BurnE2E
 {
+    using System;
     using System.Collections.Generic;
     using System.IO;
     using WixBuildTools.TestSupport;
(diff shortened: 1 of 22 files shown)


src/wix/wixnative/smartcab.cpp
```diff
@@ -1,14 +1,15 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
 
 #include "precomp.h"
+#include "fileutil.h"
```
rseanhall, line 4: I think this should have been added to `precomp.h` instead.

src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
```diff
@@ -17,23 +18,35 @@ public class CacheTests : BurnE2ETests
...
+            const long FiveGB = 5_368_709_120;
+            const long OneGB = 1_073_741_824;
             var targetFilePath = Path.Combine(this.TestContext.TestDataFolder, "fivegb.file");
+
+            // If the drive has less than 5GB (for the test file) plus 1GB (for working space), then
+            // skip the test.
+            var drive = new DriveInfo(targetFilePath.Substring(0, 1));
+            if (drive.AvailableFreeSpace < FiveGB + OneGB)
+            {
+                Console.WriteLine("Skipping CanCache5GBFile() test because there is not enough disk space available to run the test.");
```
rseanhall, line 31: The 5GB file was created while building all of the test bundles and then deleted afterwards, so there's pretty much a 0% chance of this occurring. A similar check would needed to be added to skip creating the file and building the bundle for this to be effective.

src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
```diff
@@ -17,23 +18,35 @@ public class CacheTests : BurnE2ETests
...
+                return;
```
rseanhall, line 32: Ideally we would get the test marked as skipped instead of passing.

On a related note, the move to the mono repo lost the uploading of all the bundle logs as an artifact. Once we start running these tests in CI, it will be nearly impossible to diagnose failures.

src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
robmen: I had test runs fail until I added this code. This was the last fix needed to get everything passing.

src/wix/wixnative/smartcab.cpp
robmen: Actually, I missed it when removing the CI test code I added.

src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
robmen: I wanted to mark it skipped but didn't know of a way to do that in XUnit while the test was running (admittedly I didn't look very hard last night).

Didn't mean to lose the bundle log upload. Can you provide more information (preferably in an issue so I don't lose it) and I'll fix it up? Diagnosing CI issues is challenging so I definitely want to make it easier.

src/test/burn/WixToolsetTest.BurnE2E/CacheTests.cs
rseanhall: Maybe all the debugging log statements filled up the hard drive? :)

PR #55, opened 2021-12-06
Use VS2022 build image

rseanhall:
Update v4.0 projects to target v4.7.2.
Skip all Bal managed host tests, tracked in https://github.com/wixtoolset/issues/issues/6651.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ env:
 jobs:
   build:
     name: Build
-    runs-on: windows-latest
+    runs-on: windows-2022
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ branches:
     - master
     - develop
 
-image: Visual Studio 2019
+image: Visual Studio 2022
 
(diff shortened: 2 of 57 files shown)


rseanhall:
Weird. This worked on Appveyor - https://ci.appveyor.com/project/SeanHall/wix4/builds/41790360

barnson:
Hey, this all looks familiar...I'm assuming AppVeyor's image has the reference assemblies installed (somehow).

rseanhall:
It does not, that's why I had to retarget to v4.7.2.

rseanhall:
I incorporated Bob's changes to only use v143 since the Github build machine doesn't seem to have any other toolset. Not sure if we actually want to do this or not. Or maybe start ignoring Github builds again and rely on Appveyor's until Github has the toolsets that we need?

barnson:
This means DTF can't be used by earlier targets?

rseanhall:
No, only test projects were retargeted. None of my changes affects shipped code, unless you count building it with VS2022 instead of VS2019.

barnson:
Cool.

rseanhall:
The build worked, so I guess it's time to install VS2022 locally. Still not sure what to do with this PR.

PR #57, opened 2021-12-08
5911 - BA configurable REINSTALLMODE

rseanhall:
Closes https://github.com/wixtoolset/issues/issues/5911.


src/burn/engine/logging.cpp
```diff
@@ -480,6 +478,23 @@ extern "C" LPCSTR LoggingPackageRegistrationStateToString(
...
 }
 
+extern "C" LPCSTR LoggingMsiFileVersioningToString(
+    __in BOOTSTRAPPER_MSI_FILE_VERSIONING fileVersioning
+    )
+{
+    switch (fileVersioning)
+    {
+    case BOOTSTRAPPER_MSI_FILE_VERSIONING_MISSING_OR_OLDER:
+        return "o";
```
barnson, line 488: There are constants that might look prettier: https://docs.microsoft.com/en-us/windows/win32/api/msi/nf-msi-msireinstallproductw

rseanhall: You want the value, like `4`? Or the string `"REINSTALLMODE_FILEOLDERVERSION"`? I'm not sure that either is better.

barnson: I said `might`. :) Agreed, it's not really better...

PR #61, opened 2021-12-12
Switch to `net472` as Framework baseline.

barnson:


barnson:
This also fixes a build problem I had with `net461`.

PR #63, opened 2021-12-12
Bob/5933.dependency

barnson:


rseanhall:
So the tests still failed but the build passed? That's not good.

barnson:
I concur.

rseanhall:
You renamed the column and table names in the C++ code, but only renamed the table names in the table definitions.

barnson:
Yep, easy enough fix. And hey look, the tests found a genuine bug!

PR #65, opened 2021-12-14
Configure VS with required components before building

robmen:


src/wix.vsconfig
```diff
@@ -0,0 +1,6 @@
+{
+  "version": "1.0",
+  "components": [
+    "Microsoft.Net.Component.4.7.2.TargetingPack"
```
barnson, line 4: If we're doing this, should we add the `v141` toolset/libs too? I weep for what this does to the build time...

src/vs_config.cmd
```diff
@@ -0,0 +1,11 @@
+@echo off
+setlocal enabledelayedexpansion
+
+echo Configuring Visual Studio to build WiX Toolset...
+
+for /f "usebackq tokens=*" %%i in (`"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath`) do (
+  echo Visual Studio installed at: "%%i"
+  rem "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" modify --config "%~dp0wix.vsconfig" --installPath "%%i" --quiet --norestart --force
```
rseanhall, line 8: Am I missing something? Starting the line with `rem` means the whole line is a comment so this batch file doesn't actually configure VS.

robmen: ;scream:

src/wix.vsconfig
robmen: Nope, not 0s. I had a cherry-pick failure that @rseanhall caught. It seems configuring VS takes 6+ min.

robmen:
Hmm, this was not supposed to be merged. I updated the branch. I thought GitHub said that now reset the "Auto-merge". Maybe it doesn't reset "Maintainers"? Grr. Azure DevOps is better in this regard. 😢

PR #66, opened 2021-12-14
Capture logs from burn integration tests

robmen:
Fixes wixtoolset/issues#6664

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,3 +37,9 @@ jobs:
         with:
           name: artifacts
           path: build/artifacts/
+
+      - name: Save logs
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          path: build/logs/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
         shell: cmd
         run: ./src/build_official.cmd
         env:
-          RuntimeTestsEnabled: ${{ github.ref == 'refs/heads/master' }}
+          RuntimeTestsEnabled: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/develop' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         shell: cmd
         run: ./src/build_official.cmd
         env:
-          RuntimeTestsEnabled: ${{ github.ref == 'refs/heads/master' }}
+          RuntimeTestsEnabled: true
```
rseanhall, line 31: As far as I can tell, the Github Actions build has always been running the RuntimeTests. So I guess there's an additional issue here if you didn't do that on purpose.

robmen: I checked and the `RuntimeTestsEnabled` was set to `false` correctly at this point so the issue would have been in the batch files, presumably.

rseanhall: Ok. I wanted to make sure you are aware that changing this line doesn't affect anything today.

robmen: Yeah, it's this code in `test.cmd`, if `RuntimeTestsEnabled` is set to anything tests run.

```
@if "%RuntimeTestsEnabled%"=="" echo Build integration tests %_C%
@if not "%RuntimeTestsEnabled%"=="" set _T=test&echo Run integration tests %_C%
```

I'll fix it as part of this.

rseanhall:
This doesn't quite capture what happened with Appveyor. It's very important that the logs are uploaded even if any other part of the build failed. And it's also important that the build fails if the tests fail.

rseanhall:
Also, I added the timestamp to the logs.zip filename so I didn't have to deal with `logs.zip`, `logs (1).zip`, etc in my Downloads folder. Github actions is making me just download logs.zip right now.

robmen:
I added the workflow run number to the zip of log files. You can ignore the `logs.zip`, in general. For example, in the browser use open file instead of downloading it, and grab the zip file out of it.

But would be interesting to see if I could add the workflow run number to the artifact name.

PR #67, opened 2021-12-15
6348 - Deunify FilesInUse messages in Burn

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/6348


src/api/burn/balutil/inc/IBootstrapperApplication.h
```diff
@@ -488,27 +488,19 @@ DECLARE_INTERFACE_IID_(IBootstrapperApplication, IUnknown, "53C31D56-49C0-426B-A
         __inout int* pResult
         ) = 0;
 
-    // OnExecuteFilesInUse - called when the engine encounters files in use while
-    //                       executing a package.
+    // OnExecuteFilesInUse - called when the engine receives a files in use message
+    //                       while executing a package.
     //
-    // Return:
```
barnson, line 494: Why remove these? This is the only "doc" without spelunking source.

src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
```diff
@@ -1450,6 +1445,25 @@ public enum ErrorType
...
+    public enum FilesInUseType
+    {
+        /// <summary>
+        /// Generated from INSTALLMESSAGE_FILESINUSE.
+        /// </summary>
+        Msi,
+        /// <summary>
+        /// Generated from INSTALLMESSAGE_RMFILESINUSE.
+        /// </summary>
+        MsiRm,
```
barnson, line 1460: Should we be spelling out Restart Manager (here and/or in the native code)?

src/api/burn/balutil/inc/IBootstrapperApplication.h
rseanhall: You mean source code? Burn doesn't own how the return codes are processed. The BA needs to look at the source of the FilesInUse message and return a value based on that source's documentation which is provided as a link here.

src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
rseanhall: I guess I was following Microsoft's lead here, they always used RM - `INSTALLMESSAGE_RMFILESINUSE`, `MsiRMFilesInUse` (Dialog), etc.

src/api/burn/balutil/inc/IBootstrapperApplication.h
barnson: I didn't realize it was just pass-through. Hopefully we'll never need more than an int...

src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
barnson: Yeah, that's a typical Microsoft native-code approach. Not so much with managed code, though. And Burn hasn't shied away from long-ass names...

src/api/burn/balutil/inc/IBootstrapperApplication.h
rseanhall: I guess it's not exactly pass-through, there's still some coercion by calling `UserExperienceCheckExecuteResult`. I wasn't sure whether to remove that, too.

barnson: Seems that it should be completely pass-through (though we still want to be able to talk to Burn) or translate Burn return values to the underlying engine (as Burn is hard-coded for the ones it supports).

PR #70, opened 2021-12-19
Consistent-ify primary key column names.

barnson:


barnson:
https://github.com/wixtoolset/issues/issues/5933

PR #71, opened 2021-12-21
Fix error strings that are formatted by LogErrorIdModule

rseanhall:
http://lists.wixtoolset.org/pipermail/wix-users-wixtoolset.org/2021-December/010062.html


barnson:
oof

PR #73, opened 2021-12-27
Remove RebootRequired per...

barnson:
...https://github.com/wixtoolset/issues/issues/6668.


src/burn/engine/registration.cpp
```diff
@@ -521,14 +517,6 @@ extern "C" HRESULT RegistrationDetectResumeType(
     HKEY hkRegistration = NULL;
     DWORD dwResume = 0;
 
-    if (IsBundleRebootPending(pRegistration))
-    {
-        LogId(REPORT_STANDARD, MSG_PENDING_REBOOT_DETECTED, pRegistration->sczRegistrationKey);
-
-        *pResumeType = BOOTSTRAPPER_RESUME_TYPE_REBOOT_PENDING;
```
rseanhall, line 528: We should also remove `BOOTSTRAPPER_RESUME_TYPE_REBOOT_PENDING`.

barnson: D'oh. Thought I caught that one.

PR #74, opened 2021-12-27
Add semver to help command

MarkStega:
Fixes wixtoolset/issues#6474

### DIFF
--- a/src/wix/WixToolset.Core/CommandLine/CommandLine.cs
+++ b/src/wix/WixToolset.Core/CommandLine/CommandLine.cs
@@ -56,7 +56,7 @@ namespace WixToolset.Core.CommandLine
             if (command.ShowLogo)
             {
                 var branding = this.ServiceProvider.GetService<IWixBranding>();
-                Console.WriteLine(branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version [FileVersion]"));
+                Console.WriteLine(branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version [ProductVersion]"));
                 Console.WriteLine(branding.ReplacePlaceholders("[AssemblyCopyright]"));
             }
 
--- a/src/wix/WixToolset.Core/CommandLine/HelpCommand.cs
+++ b/src/wix/WixToolset.Core/CommandLine/HelpCommand.cs
@@ -54,7 +54,7 @@ namespace WixToolset.Core.CommandLine
             }
 
             Console.WriteLine();
-            Console.WriteLine("Run 'wix [command] --help' for more information on a command.");
+            Console.WriteLine("Run 'wix [command] -h[elp]' for more information on a command.");
             Console.WriteLine();
             Console.WriteLine(this.Branding.ReplacePlaceholders("For more information see: [SupportUrl]"));
 
--- a/src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
+++ b/src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
@@ -22,7 +22,6 @@ namespace WixToolset.Core.CommandLine
                                                   , ThisAssembly.Git.BaseVersion.Patch
                                                   , ThisAssembly.Git.SemVer.DashLabel
                                                   , ThisAssembly.Git.Commit);
-
             return Task.FromResult(0);
         }
 
--- a/src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
+++ b/src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
@@ -61,6 +61,7 @@ namespace WixToolset.Core.ExtensibilityServices
 
                 original = original.Replace("[FileCopyright]", commonFileVersion.LegalCopyright);
                 original = original.Replace("[FileVersion]", commonFileVersion.FileVersion);
+                original = original.Replace("[ProductVersion]", commonFileVersion.ProductVersion);
             }
 
             var fileVersion = FileVersionInfo.GetVersionInfo(assembly.Location);
@@ -69,6 +70,7 @@ namespace WixToolset.Core.ExtensibilityServices
             original = original.Replace("[FileCopyright]", fileVersion.LegalCopyright);
             original = original.Replace("[FileProductName]", fileVersion.ProductName);
             original = original.Replace("[FileVersion]", fileVersion.FileVersion);
+            original = original.Replace("[ProductVersion]", fileVersion.ProductVersion);
 
             if (original.Contains("[FileVersionMajorMinor]"))
             {


MarkStega:

MarkStega:
This PR was done (mostly) to check that the VS2022 build of wix was correct on my development system. It will close out issue #6474.

I simply replicated the write of the semantic version from the version command into the help command.

I submitted as draft because it might make more sense to add the semver to the 'standard header' created in commandline.cs lines 59 & 60.

robmen:
Close but the goal should be to make this line:

```
WiX Toolset Core version 4.0.0.0
```

Look more like this:

```
WiX Toolset Core version 4.0.0-preview.0+824e72cf
```

That way the WiX Toolset header will look more like MSBuild header.

rseanhall:
I assume @MarkStega was just following the expected output that was described in https://github.com/wixtoolset/issues/issues/6474. If we go your way, then it sounds like that line should be removed from the version command.

robmen:
The `version` command is intended to print out the version of the WiX Toolset on a line with no adornments before or after it.

MarkStega:
I was following the suggested output in [#6474](https://github.com/wixtoolset/issues/issues/6474). I'm a little confused because the version command currently outputs 3 lines:
```
WiX Toolset Core version 4.0.0.121
Copyright (c) .NET Foundation and contributors. All rights reserved.
4.0.0-preview.1-build.121+4c34ad46
```
The first two lines are in common to version & help commands. If I adorn the first line then I'd suggest that the version command output command should be
```
WiX Toolset Core version 4.0.0-preview.1-build.121+4c34ad46
Copyright (c) .NET Foundation and contributors. All rights reserved.
4.0.0.121
```
I can swizzle the text however you want, I just need some clarity of the desired output of the two headers (version & help).

robmen:
I updated the original issue.

I hadn't updated it because the issue was assigned to me and had a rough idea of what I was looking for.

PS: in the future, leave a comment on an issue that you'd like to tackle so:
1. It can be assigned to you and multiple people don't try to solve the same issue.
2. If it is already assigned to someone else, that person can provide any context he or she has not yet transferred into the issue text.

MarkStega:
Rob,

Thanks, I'll make certain that I do a note in the future.

So do you want a third line from the version command or should it simply be the same as the help command?

robmen:
> The `version` command is intended to print out the version of the WiX Toolset on a line with no adornments before or after it.

So the `version` command should not change. The change is to print the full semver in the header (for every command).

barnson:
So `wix version -nologo` prints one line, just the unadorned version?

MarkStega:
@robmen If the header has the semver then if not changed the version command output would be
```
WiX Toolset Core version 4.0.0-preview.1-build.121+4c34ad46
Copyright (c) .NET Foundation and contributors. All rights reserved.
4.0.0-preview.1-build.121+4c34ad46
```
So I'd suggest removing the third output line which has the second semver.

@barnson Adding -nologo doesn't change the output from the version command.

robmen:
@barnson Yes, exactly.

@MarkStega:
>  Adding -nologo doesn't change the output from the version command.

Well, it is supposed to. `-nologo` should always prevent the header from printing.

MarkStega:
Here is where the output stands now for help with the semver in the header (I also fixed the 'more information' line):
```
C:\Solutions\OHI\Optimiser\WiX.2012>C:\Solutions\TemporarySolutions\wix4\build\wix\Release\publish\wix\wix --help
WiX Toolset Core version 4.0.0-preview.1-build.122+5b03b216
Copyright (c) .NET Foundation and contributors. All rights reserved.

Usage: wix [option]
Usage: wix [command]

Options:
  -h|--help         Show command line help.
  --version         Display WiX Toolset version in use.

Commands:
  build             Build a wixlib, package or bundle.
  decompile         Decompile a package or bundle into source code.
  convert           Convert v3 source code to v4 source code.
  extension         Manage extension cache.
  format            Ensures consistent formatting of source code.

Run 'wix [command] -h[elp]' or 'wix [command] -?'for more information on a command.

For more information see: http://wixtoolset.org/
```
And the version option:
```
C:\Solutions\OHI\Optimiser\WiX.2012>C:\Solutions\TemporarySolutions\wix4\build\wix\Release\publish\wix\wix --version
WiX Toolset Core version 4.0.0-preview.1-build.122+5b03b216
Copyright (c) .NET Foundation and contributors. All rights reserved.
```
I didn't try to fix the -nologo issue as that was independent of this work & pre-existed. I am reasonably certain that the issue revolves around FixupCommandBase.cs

Is the expectation such that the header should be shown for all commands (unless suppressed by '-nologo')? I ask because 'wix convert' and 'wix build' commands do not show the logo.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
```diff
@@ -16,13 +16,6 @@ internal class VersionCommand : ICommandLineCommand
 
         public Task<int> ExecuteAsync(CancellationToken cancellationToken)
         {
-            // $(GitBaseVersionMajor).$(GitBaseVersionMinor).$(GitBaseVersionPatch)$(GitSemVerDashLabel)+$(Commit)
-            Console.WriteLine("{0}.{1}.{2}{3}+{4}", ThisAssembly.Git.BaseVersion.Major
-                                                  , ThisAssembly.Git.BaseVersion.Minor
-                                                  , ThisAssembly.Git.BaseVersion.Patch
-                                                  , ThisAssembly.Git.SemVer.DashLabel
-                                                  , ThisAssembly.Git.Commit);
-
```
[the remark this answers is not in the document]
MarkStega: If this isn't removed then the console output has a redundant line which just repeats information from the logo (first two lines):
```
WiX Toolset Core version 4.0.0-preview.1-build.122+5b03b216
Copyright (c) .NET Foundation and contributors. All rights reserved.
4.0.0-preview.1-build.122+5b03b216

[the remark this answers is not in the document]
Comment (robmen) [reply src/wix/WixToolset.Core/CommandLine/VersionCommand.cs:25]: If `-nologo` is not provided, that is correct. The header should be 💯 controlled by the `-nologo` switch (at least, until we choose to change the header behavior and take into account all the things @barnson raised above).

File: src/wix/WixToolset.Core/CommandLine/HelpCommand.cs [hunk 2]
```diff
@@ -54,7 +54,7 @@ public Task<int> ExecuteAsync(CancellationToken _)
             }
 
             Console.WriteLine();
-            Console.WriteLine("Run 'wix [command] --help' for more information on a command.");
+            Console.WriteLine("Run 'wix [command] -h[elp]' or 'wix [command] -?'for more information on a command.");
```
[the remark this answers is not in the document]
MarkStega: Done...

src/wix/WixToolset.Core/CommandLine/CommandLine.cs
```diff
@@ -56,7 +56,14 @@ public ICommandLineCommand ParseStandardCommandLine(ICommandLineArguments argume
...
                 var branding = this.ServiceProvider.GetService<IWixBranding>();
-                Console.WriteLine(branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version [FileVersion]"));
+                // WiX Toolset Core version $(GitBaseVersionMajor).$(GitBaseVersionMinor).$(GitBaseVersionPatch)$(GitSemVerDashLabel)+$(Commit)
+                Console.WriteLine(
+                    branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version {0}.{1}.{2}{3}+{4}")
+                    .Replace("{0}", ThisAssembly.Git.BaseVersion.Major)
+                    .Replace("{1}", ThisAssembly.Git.BaseVersion.Minor)
+                    .Replace("{2}", ThisAssembly.Git.BaseVersion.Patch)
+                    .Replace("{3}", ThisAssembly.Git.SemVer.DashLabel)
+                    .Replace("{4}", ThisAssembly.Git.Commit));
```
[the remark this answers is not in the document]
MarkStega: Done...

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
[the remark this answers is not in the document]
MarkStega: For now, since the switches are ignored in the help & version options I left it as is... Easy enough to add when the switches are supported.

MarkStega:
BTW, I am not going to even pretend that I understand the WiX buil at this point but there is a bit of redundancy that I don't understand. Much (?All?) of the code in WixBranding.cs is replicated in WixDistribution.cs; Does the version mod need to reside there also?

rseanhall:
I can go into the history if you want, but the short version is that you can ignore src\test\burn\WixToolset.WixBA\WixDistribution.cs.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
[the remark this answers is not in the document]
rseanhall: This is what I get when I run the version command for MSBuild:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise>msbuild -version
Microsoft (R) Build Engine version 17.0.0+c9eb9dd64 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

17.0.0.52104
```

You'll notice that there are two versions here, too. That's what we want, even if it's redundant.

src/wix/WixToolset.Core/CommandLine/CommandLine.cs
```diff
@@ -56,7 +56,7 @@ public ICommandLineCommand ParseStandardCommandLine(ICommandLineArguments argume
             if (command.ShowLogo)
             {
                 var branding = this.ServiceProvider.GetService<IWixBranding>();
-                Console.WriteLine(branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version [FileVersion]"));
+                Console.WriteLine(branding.ReplacePlaceholders("[AssemblyProduct] [AssemblyDescription] version [AssemblyVersion]"));
```
robmen, line 59: I probably should have expounded upon this comment a bit more:

> Better to update WixBranding to add a new ProductVersion replacement and use that. Then this won't be a special casing using ThisAssembly.

Rather than changing the behavior of `AssemblyVersion` add a new `ProductVersion` replacement to WixBranding. Use `ProductVersion` here instead of `FileVersion` and (now) `AssemblyVersion`.

src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
```diff
@@ -101,6 +102,20 @@ public string ReplacePlaceholders(string original, Assembly assembly = null)
...
+                    ThisAssembly.Git.BaseVersion.Major + "." +
+                    ThisAssembly.Git.BaseVersion.Minor + "." +
+                    ThisAssembly.Git.BaseVersion.Patch +
+                    ThisAssembly.Git.SemVer.DashLabel + "-" +
+                    ThisAssembly.Git.Branch + "+" +
+                    ThisAssembly.Git.Commit;
+
+                original = original.Replace("[AssemblyVersion]", assemblyVersion);
+            }
+            catch { }
```
robmen, line 117: I probably should have expounded upon this comment a bit more:

> Better to update WixBranding to add a new ProductVersion replacement and use that. Then this won't be a special casing using ThisAssembly.

We should avoid doing this string concatenation that is already done in the build process to create the versions in the files themselves. To ensure the values displayed here match the values in the files, let's use the file version information in the files themselves.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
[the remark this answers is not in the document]
MarkStega: East enough -- Its major.minor.patch.'something else' What is the 'something else' from GitInfo?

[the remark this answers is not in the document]
robmen: Interestingly, MSBuild prints out two different versions. Let's not worry about that for now and use the `ProductVersion` as the version displayed for WiX.

src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
robmen: To do so, add lines near 64 and 72 to do the `[ProductVersion]` replacement much the same as the `[FileVersion]` replacement is done today.

robmen (changes requested):

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
robmen, line 25: Isn't this removing the output of the `version` command? I think we still want this part, right?

src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
```diff
@@ -10,6 +10,7 @@ namespace WixToolset.Core.ExtensibilityServices
     using System.Diagnostics;
     using System.IO;
     using System.Reflection;
+
```
robmen, line 13: Why add whitespace here?

robmen (changes requested):

robmen:
PS: Please squash your changes down to a single commit. We don't need to see the previous steps fully exploring this space. 😄

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
MarkStega: I don't think so -- The version is shown in the logo and since the version command explicitly ignores the -nologo switch it would be redundant:
```
C:\Solutions\OHI\Optimiser\WiX.2012>C:\Solutions\TemporarySolutions\wix4\build\wix\Release\publish\wix\wix --version
WiX Toolset Core version 4.0.0-preview.1-build.126+4d48856ec445e3ba2efa256e166f01455a859ff1
Copyright (c) .NET Foundation and contributors. All rights reserved.

File: src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs [hunk 7]
```diff
@@ -101,6 +104,20 @@ public string ReplacePlaceholders(string original, Assembly assembly = null)
...
+            //        ThisAssembly.Git.BaseVersion.Minor + "." +
+            //        ThisAssembly.Git.BaseVersion.Patch +
+            //        ThisAssembly.Git.SemVer.DashLabel + "-" +
+            //        ThisAssembly.Git.Branch + "+" +
+            //        ThisAssembly.Git.Commit;
+
+            //    original = original.Replace("[AssemblyVersion]", assemblyVersion);
+            //}
+            //catch { }
+
```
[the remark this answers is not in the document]
MarkStega: Correct - I left it in just in case 'productversion' wasn't what was wanted.

src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
MarkStega: That was a result of VS doing a 'remove unused usings and sort'. It's gone.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
robmen: Sorry, let me be more clear. We want it. It is (always right now without `-nologo` support) redundant but we want it.

Fixing the issue does not require updating the `version` command at all.

MarkStega:
> PS: Please squash your changes down to a single commit. We don't need to see the previous steps fully exploring this space. 😄

Interesting thought -- I've always followed the pattern of accepting the PR with a 'Squash and Merge' but I can certainly see the benefit of keeping the PR 'clean' so I went ahead and quashed the commits.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
MarkStega: Would it be better to use the same version output as the logo instead of using the ThisAssembly.Git information?

robmen:
I squash by doing a `git rebase -i develop` (when in the branch I want to squash) then fixup the commits to best represent what I was trying to do.

Admittedly, I sometimes get too much unrelated stuff on my machine and don't always get the commits separated. But in cases like this, its definitely better to clean up the history (unless the path to get to the solution provided real value).

rseanhall:
If the PR will squash down to one commit, we might want to consider doing the squash for them in the Github UI since it regularly trips people up. I did this with one of Nir's PRs.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
robmen: No. At this time, the `version` command always provides the full git version details. The `ProductVersion` will contain less information in "official" builds (for example the SHA hash is removed).

src/wix/WixToolset.Core/ExtensibilityServices/WixBranding.cs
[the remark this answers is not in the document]
robmen: Yeah, won't need this in the final commit.

robmen:
@MarkStega if you want me to do the squash for you, I can do that. If you want the experience of "force pushing" your squashed commits, that's cool as well.

MarkStega:
@robmen - After a 'breaking' interactive rebase I figured out that I had to do the rebase followed by a force push. So there is now 1 commit with the very modest changes.

@rseanhall -- Yes, a 'squash & merge' acceptance of a PR forces the history down to a single commit, but since I fought my way through a successful rebase on my second try it shouldn't be needed.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
```diff
@@ -14,15 +14,19 @@ internal class VersionCommand : ICommandLineCommand
...
-                                                  , ThisAssembly.Git.BaseVersion.Patch
-                                                  , ThisAssembly.Git.SemVer.DashLabel
-                                                  , ThisAssembly.Git.Commit);
+            this.ServiceProvider = serviceProvider;
+        }
 
+        public Task<int> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            var branding = this.ServiceProvider.GetService<IWixBranding>();
+            Console.WriteLine(branding.ReplacePlaceholders("[ProductVersion]"));
```
robmen, line 29: I think [my reply above](https://github.com/wixtoolset/wix4/pull/74/files#r778878727) got lost in all the other communication.

For this fix, there shouldn't be any changes to the `VersionCommand.cs` .

robmen (changes requested):
This does look like it will be a pretty small change since the `version` command isn't being updated.

robmen:
@MarkStega glad you figured the rebase/force push dance out. It's a great git technique to have in your toolbox. The single commit represents the change appropriately (even if it misses out on the journey to get there). Looking forward to getting this one in. 👍🏻

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
MarkStega: Rob, sorry I missed this comment

I am a bit confused (as always? :-) ) but in my 'official build' here are the two outputs, PV is the ProductVersion info and GI is the github info version:
```
C:\Solutions\OHI\Optimiser\WiX.2012>C:\Solutions\TemporarySolutions\wix4\build\wix\Release\publish\wix\wix --version
WiX Toolset Core version 4.0.0-preview.1-build.162+0fb2a3c27c5ca58a28dae07053ad1411735b0889
Copyright (c) .NET Foundation and contributors. All rights reserved.
PV: 4.0.0-preview.1-build.162+0fb2a3c27c5ca58a28dae07053ad1411735b0889
GI: 4.0.0-preview.1-build.162+0fb2a3c2
```
I'll implement the GI version, just trying to understand your comment. When I run the 'build_official command I take it that the result is different then the release build?

robmen: No worries. I missed your comment above mine for quite a while as well.

My memory of the official build is faulty. In `src\Directory.Build.targets` the `InformationalVersion` (aka: ProductVersion) is essentially the same as the value displayed in the `version` command. My thinking is still that the `version` command should display "all the information available at build time" and the header should display "whatever the product version of the file is". If/when those versions change then the code will remain correct.

My faulty memory is that I thought the ProductVersion already differed.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
```diff
@@ -14,15 +14,17 @@ internal class VersionCommand : ICommandLineCommand
...
-                                                  , ThisAssembly.Git.BaseVersion.Patch
-                                                  , ThisAssembly.Git.SemVer.DashLabel
-                                                  , ThisAssembly.Git.Commit);
-
+            Console.WriteLine(
+                ThisAssembly.Git.BaseVersion.Major + "." +
+                ThisAssembly.Git.BaseVersion.Minor + "." +
+                ThisAssembly.Git.BaseVersion.Patch +
+                ThisAssembly.Git.SemVer.DashLabel + "+" +
+                ThisAssembly.Git.Commit);
```
robmen, line 27: Why switch to a bunch of string concatenations instead of letting `Console.Writeline()` handle the single string format internally? The compiler _probably_ figures out all those string concatenations can be turned into a `String.Concat()` call but it doesn't seem worth it to test the compiler this way. It won't matter in this case but avoiding string allocations is generally a good idea. Is there some reason this is better?

robmen (approved):
I'm approving this but I am curious what your thinking was for the change in the `version` command.

src/wix/WixToolset.Core/CommandLine/VersionCommand.cs
MarkStega: I just thought this was a better representation of the desired output. It's a personal choice (because I don't like 'magic'/'I don't know how this is implemented in the writeline case') so if you prefer that style I can easily revert.

robmen: Yeah. I would recommend that.

MarkStega: No problem, done

robmen:
@MarkStega Awesome. Thank you for working through all the discussion and squashing. If you want me to assign wixtoolset/issues#6474 to you (for credit?) leave a comment on the issue and I'll assign it to you (could have done this a while ago 😄 ).

And if you're interested in continuing to help out, let me know and we'll see what other issues might be interesting to you.

PR #77, opened 2021-12-28
More build improvements

robmen:


src/internal/SetBuildNumber/Directory.Packages.props.pp
```diff
@@ -42,6 +42,7 @@
     <PackageVersion Include="System.Text.Encoding.CodePages" Version="4.6.0" />
 
     <PackageVersion Include="Microsoft.AspNetCore.Owin" Version="3.1.13" />
+    <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" />
```
rseanhall, line 45: Just remove this and the references to this package everywhere. They're only causing problems with no benefit.

robmen: I really wanted to understand what was going on here... but I give up.

src/Directory.Build.rsp
```diff
@@ -0,0 +1,4 @@
+-nologo
+-nr:false
+-v:m
```
rseanhall, line 3: 1. When are these used?
2. Isn't -nologo redundant? It seems to be specified everywhere already.
3. How do you get the binlogs from the CI server?

src/internal/SetBuildNumber/Directory.Packages.props.pp
rseanhall: I assume we're hitting another edge case in Nuget that no one else cares about.

src/Directory.Build.rsp
robmen: 1. Every build.
2. There is a bug open in MSBuild to respect `-nologo` in the response files. So for now it has to be added everywhere.
3. The "Summary" of the build shows all the uploaded artifacts. `Logs_#####` is one of them. You can "Open" that then double click the .binlogs inside if you have binlog viewer installed. Better experience than I expected.

src/internal/SetBuildNumber/Directory.Packages.props.pp
robmen: Yeah, CPVM is still "preview" so part of it is my own doing but I was hoping I could understand the root issue to possibly better handle something like it in the `wix.targets` in the future.

But you're probably right that it is a collection of edge cases no one really cares about.

src/Directory.Build.rsp
rseanhall: I feel like we need binlogs for everything if everything is building with minimal verbosity.

robmen: We need binlogs for all builds.

PR #80, opened 2021-12-31
Re-enable test fixed by wix.targets rewrite

robmen:
Closes wixtoolset/issues#6407

### DIFF
--- a/src/wix/test/WixToolsetTest.Sdk/MsbuildFixture.cs
+++ b/src/wix/test/WixToolsetTest.Sdk/MsbuildFixture.cs
@@ -44,7 +44,6 @@ namespace WixToolsetTest.Sdk
             }
         }
 
-        [Theory(Skip = "Test demonstrates failure")]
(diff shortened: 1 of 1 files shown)


src/wix/test/WixToolsetTest.Sdk/MsbuildFixture.cs
```diff
@@ -44,7 +44,6 @@ public void CanBuildSimpleBundle(BuildSystem buildSystem)
             }
         }
 
-        [Theory(Skip = "Test demonstrates failure")]
```
rseanhall, line 47: I'm pretty sure the Theory` attribute is required for it to be recognized as a test, I would be surprised if this is actually running.

robmen: Very possibly. I don't use `Theory` because the error reporting is difficult to diagnose. I'm retesting.

PR #82, opened 2022-01-02
Actually fix uncompressed bundle payloads

robmen:
Fixes wixtoolset/issues#6407


src/create_zips.cmd
```diff
@@ -1,2 +1,2 @@
-7z a "build\logs\test_burn_%GITHUB_RUN_ID%.zip" "%TEMP%\*.log" "%TEMP%\..\*.log" || exit /b
-7z a "build\testresults.zip" @src\testresultfilelist.txt || exit /b
\ No newline at end of file
+7z a "build\logs\test_burn_%GITHUB_RUN_ID%.zip" "%TEMP%\*.log" "%TEMP%\..\*.log"
+7z a "build\testresults.zip" @src\testresultfilelist.txt
```
rseanhall, line 2: Nope, that's the whole point of creating this zip. The build needs to fail if tests weren't run for some reason, like someone messed up a build script.

PR #83, opened 2022-01-03
Add nupkg push

robmen:


.github/workflows/build.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           name: artifacts
           path: build/artifacts/
 
-      - name: Zip temp logs
+      - name: Collect integration test logs
```
rseanhall, line 51: FYI, they're collecting the logs from the Configure Visual Studio step as well.

.github/workflows/build.yml
```diff
@@ -59,3 +59,6 @@ jobs:
         with:
           name: logs_${{ github.run_id }}
           path: build/logs/
+
+      - name: Push to GitHub Packages
+        run: dotnet nuget push "build/artifacts/**/*.nupkg" --source https://nuget.pkg.github.com/wixtoolset/index.json --api-key ${{ secrets.GITHUB_TOKEN }} --skip-duplicate
```
rseanhall, line 64: This probably needs to be restricted to not run in PR builds and not run in forked repos.

robmen: Yeah, that's the last step. It is very possible this go around will fail because one or more .nupkgs will need their `RepositoryUrl` set properly.

.github/workflows/build.yml
robmen: That is true (plus any other unrelated *.log files). I've only ever used the log files to debug burn integration test files. The name `test_burn_%GITHUB_RUN_ID%.zip` suggests that is its primary purpose. Would it be clearer if the name of the .zip file was changed? Given I've only used it for debugging burn integration tests, it makes sense to me right now but I'd be interested if other names are more appropriate.

rseanhall: Probably. I'm still annoyed at having to extract a zip file from a zip file, but I learned that that is the best way to upload a bunch of files with Github Actions right now.

.github/workflows/build.yml
rseanhall: Is this resolved? What's going to happen when I sync my repo and the build automatically kicks off? My build fails because it doesn't have rights to push to the `wixtoolset` feed?

robmen: I did take the repository_owner out of the URL so I now added it to the condition.

PR #84, opened 2022-01-03
Replace create_zips.cmd with Github Action steps

rseanhall:


.github/workflows/build.yml
```diff
@@ -44,9 +44,13 @@ jobs:
           name: artifacts
           path: build/artifacts/
 
-      - name: Create Zips
+      - name: Validate Test Results
         shell: cmd
-        run: ./src/create_zips.cmd
+        run: 7z a build\testresults.zip @src\testresultfilelist.txt
```
robmen, line 49: I believe you'll want `always()` here so it runs even if previous steps fail .

rseanhall: The zip itself is not important. The test result files are in the `build\logs` folder which get uploaded with everything else in the later step.

rseanhall: In the future, maybe someone will spend some time to replace this command with actually inspecting the results and ensuring that no tests failed.

PR #87, opened 2022-01-04
Some minor misc build fixes

robmen:


src/test/burn/test_burn.cmd
```diff
@@ -11,7 +11,7 @@
 
 @echo Burn integration tests %_C%
 
-msbuild -t:Build -Restore -p:Configuration=%_C% -warnaserror -m -bl:%_L%\test_burn_build.binlog || exit /b
```
rseanhall, line 14: I'm surprised that you were willing to spend all that time on that nuget error with referencing `Microsoft.NETFramework.ReferenceAssemblies` but didn't want to spend time getting the parallel build working here. Does that mean you're confident that the wix.targets are not to blame?

If you don't want to spend time getting it to work as is, it should be straightforward to build the .wixlibs first and then build all of the TestData after that.

src/samples/thmviewer/thmviewer.vcxproj
```diff
@@ -66,4 +66,4 @@
   </ItemGroup>
 
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
rseanhall, line 69: You should have removed line 50.

Also, why did you modernize the .wixproj when it isn't even getting built?

robmen: Still working on that part.

src/test/burn/test_burn.cmd
robmen: Pretty confident. The issue I was trying to avoid occurred during the [TestBA.csproj build](https://github.com/wixtoolset/wix4/runs/4695269193?check_suite_focus=true#step:4:2596).

rseanhall: I assumed that the errors there meant that TestBA.csproj was being built concurrently by multiple different .wixproj, when it should have deduplicated that and made all the .wixproj wait on one single instance of building the .csproj.

robmen: I could be wrong but it seemed to be caused by [this command](https://github.com/wixtoolset/wix4/runs/4695269193?check_suite_focus=true#step:4:2582):

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\msbuild.exe -nologo -bl:D:\a\wix4\wix4\src\test\burn\..\..\..\build\logs\test_burn_build.binlog -m -nr:false -p:Configuration=Release -Restore -t:Build -v:m -warnaserror .\BurnE2ETests.sln
```

Which seemed to only build .csprojs.

rseanhall: My bad, you're right. MSB3026 is supposed to be a warning, we could go back to parallel if we can create a special case for that not to be upgraded to an error.

robmen: No worries. If this warning-to-error is a widespread problem preventing multi-proc elsewhere then maybe special casing is worth it. Not sure if it is for this case as the projects seem pretty shallow and small.

PR #89, opened 2022-01-06
Always run upgrade related bundles last

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/5128


src/Directory.Build.props
```diff
@@ -6,6 +6,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <EnableSourceLink Condition=" '$(NCrunch)' == '1' ">false</EnableSourceLink>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <NoWarn>$NoWarn;MSB3026</NoWarn>
```
robmen, line 9: Is `$NoWarn` supposed to be `$(NoWarn)` to allow other props to add warnings together? Or is this a special syntax for `NoWarn` that does something cool?

rseanhall: No, that's a typo. :)

PR #91, opened 2022-01-10
Modernize signing preparation tooling

robmen:
Fixes wixtoolset/issues#4536


src/wix/WixToolset.Core.Burn/BurnExtensionCommandLine.cs
```diff
@@ -0,0 +1,41 @@
...
+        public BurnExtensionCommandLine(IServiceProvider serviceProvider)
+        {
+            this.ServiceProvider = serviceProvider;
+        }
+
+        private IServiceProvider ServiceProvider { get; }
+
+        public override IReadOnlyCollection<ExtensionCommandLineSwitch> CommandLineSwitches => new ExtensionCommandLineSwitch[]
+        {
+            new ExtensionCommandLineSwitch { Switch = "msi", Description = "Windows Installer specialized operations." },
```
barnson, line 26: `msi`?

src/wix/WixToolset.Core.Burn/BurnCommand.cs
```diff
@@ -0,0 +1,78 @@
...
+                return false;
+            }
+
+            return this.Subcommand.TryParseArgument(parser, argument);
+        }
+
+        private static void DisplayHelp()
+        {
+            Console.WriteLine();
+            Console.WriteLine("Usage: wix burn detach|reattach bundle.exe -out engine.exe");
```
barnson, line 66: Usage smells wrong, as `-out` isn't the engine, and it's missing some args, right?

robmen: This isn't complete. I have an issue I'll open when this is ready to do a sweep through the whole help system. I've got a better feel for how I want things fit together now than when I originally wrote a lot of this cli code. There is a lot of "TBW" help strings.

src/wix/WixToolset.Core.Burn/BurnExtensionCommandLine.cs
robmen: Yeah, `msi` right now. Other options: `wi` or `windowsinstaller` (which feels way too long). Thoughts?

barnson: `msi` in **Burn**ExtensionCommandLine.cs?

src/wix/WixToolset.Core.Burn/BurnCommand.cs
barnson: Fair. This comment was originally bigger -- before I saw that `-engine` was consistent across subcommands. (Originally I suggested adding `-bundle` to clarify but then decided `-out` made sense for the new bundle.)

src/wix/WixToolset.Core.Burn/BurnExtensionCommandLine.cs
robmen: Oh, duh. Fixed.

src/wix/WixToolset.Core.Burn/BurnCommand.cs
robmen: Yeah, I'm not married to these but after messing around a bit I think they made the most sense as I worked. Certainly made sense when I was writing the MSBuild task so that's a plus.

PR #92, opened 2022-01-11
Develop vs2022

cpuwzd:
These are the changes I made on my machine to build the develop branch under vs 2022.

Note most importantly that I have disabled source checking for all C# projects. These errors, if not ignored or corrected, prevent significant portions of the code from being built or tested. At some point, this issue needs to be resolved.


barnson:
😕? My changes for VS22 compatibility are already in `develop`. What problems were you seeing? The use of `v141` for libraries was intentional, to provide compatibility back to VS2017.

PR #102, opened 2022-01-30
Add WixUI compiler extension to handle platforms.

barnson:
Custom actions to print EULA and validate install directories are
defined in WixUIExtension compiler extension, to handle
platform-specific custom actions referred to from `DoAction` control
events. This is the least-worst solution, given the `DoAction` approach
used in the WixUI authoring and anyone customizing a WixUI set.


barnson:
https://github.com/wixtoolset/issues/issues/6612

PR #111, opened 2022-02-08
Add runtime testing for #6705

rseanhall:


rseanhall:
The builds are getting automatically canceled and I can't tell why.

robmen:
So, given the progress, I guess the VS install either needs to restart the computer or our configuration step isn't actually waiting for the install to complete.

rseanhall:
I'm not sure. The initial steps are still taking way too long. I reenabled AppVeyor for now.

robmen:
Oh, I definitely spoke to soon. Given the way my week is going... not surprising. :(

rseanhall:
My hypothesis build failed because they still don't have the v141 toolset for us. It might have completed successfully, but it feels like the machine is very low on either memory or disk space considering how long it takes for things that should be fast like running the VS initialization script.

rseanhall:
Looks like AppVeyor is rebasing or merging my branch with the latest, which is failing due to problems fixed in Rob's PR. We should probably figure out how to get Github Actions to do that, though I had hoped to use that fact that it doesn't to my advantage here.

rseanhall:
Close and reopen to trigger Appveyor.

rseanhall:
Appveyor build passed: https://ci.appveyor.com/project/wixtoolset/wix4/builds/42515706

PR #114, opened 2022-02-10
Central caDecor and caErr.

barnson:


barnson:
Overriding because of GitHub Actions build failures.

PR #119, opened 2022-02-18
Code base issue

cpuwzd:
The issue to which I refer as the CodeBase Issue, has to do with multiple places within the Wix 4 product that, for one reason or another, attempt to determine the file from which the currently executing code was loaded. One of the situations in which this might occur includes a vast number of system-level tests with names that begin with "CanBuild". This  can happens when the test needs to find wixnative.exe or WixToolset.Sdk.props.  Here, we are generally looking at a system that has built the Wix 4 product, but there might be scenarios in which these tests might be run during or after the installation of Wix 4 on a machine that is, for our purposes, only used to create installation packages and never used to build Wix 4, itself. Such tests might be run from the command line or from Visual Studio.
Beyond these tests, there might be other scenarios in which the Wix 4 product wants to know from where it was loaded. There are a couple of problems with trying to determine how a particular piece of code got loaded into the current process. The current code relies on either on a .NET Framework 4.72 API or a .NET Core 3.1 API, neither of which is available in .NET Core 5.0 or later. In .NET Core 5.0 and later, you can't dpend on much that is operating-system dependent.
We need a classification of all the scenarios in which this situation arises that recognizes the actual problem that is currently being solved (or perhaps just being attacked) by using knowledge of the image source. Then we need a way to provide the necessary knowledge where it is needed.
When tests are run from disposable file systems, the actual path to the loaded image is typically in the temp directory of the user account running the test. This is very different from a non-test situation.
I'm sure that I don't grasp all of the issues involved, but I have been looking at this issue (among others) for at least nine months. I think that it is time for some guidance from and decision making by those holding the keys to the realm.

### DIFF
--- a/src/ext/Bal/test/WixToolsetTest.Bal/BalExtensionFixture.cs
+++ b/src/ext/Bal/test/WixToolsetTest.Bal/BalExtensionFixture.cs
@@ -12,7 +12,7 @@ namespace WixToolsetTest.Bal
 
     public class BalExtensionFixture
     {
-        [Fact]
+        [Fact(Skip = "xxxxx CodeBase Issue: We can't determine the file path from which we were loaded. xxxxx")]
         public void CanBuildUsingDisplayInternalUICondition()
         {
             using (var fs = new DisposableFileSystem())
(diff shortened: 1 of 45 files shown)


robmen:
I'm a little confused. I know finding the Codebase mexhanism changed in .NET 5 but everything is working correctly right now. Why do we need this change?

PR #123, opened 2022-02-21
Turn off vsconfig now that everything is in the image

robmen:


rseanhall:
We seem to be getting the new image now.

PR #126, opened 2022-02-23
Fix wixlib, msm, msi incremental build

robmen:
Fixes wixtoolset/issues#4048
Fixes wixtoolset/issues#4049
Fixes wixtoolset/issues#4060


src/test/wix/TestData/WixprojPackageVcxprojWindowsApp/WixprojPackageVcxprojWindowsApp.wixproj
```diff
@@ -5,5 +5,7 @@
     <ProjectReference Include="..\VcxprojWindowsApp\VcxprojWindowsApp.vcxproj" />
     <ProjectReference Include="..\WixprojLibraryVcxprojDll\WixprojLibraryVcxprojDll.wixproj" />
     <ProjectReference Include="..\WixprojModuleCsprojWinFormsNetFx\WixprojModuleCsprojWinFormsNetFx.wixproj" />
+
+    <PackageReference Include="Wixtoolset.UI.wixext" Version="4.0.0-preview.1-build.205" />
```
rseanhall, line 9: I hope the build fails because of this hardcoded version.

src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
```diff
@@ -613,5 +618,26 @@ private string ResolveMedia(MediaSymbol media, string mediaLayoutDirectory, stri
 
             return layout;
         }
+
+        private IEnumerable<ITrackedFile> TrackInputFiles(WindowsInstallerData data, List<ITrackedFile> trackedFiles)
+        {
+            var trackedInputFiles = new List<ITrackedFile>();
+            var intermediateAndTemporaryPaths = new HashSet<string>(trackedFiles.Where(t => t.Type == TrackedFileType.Intermediate || t.Type == TrackedFileType.Temporary).Select(t => t.Path), StringComparer.OrdinalIgnoreCase);
```
rseanhall, line 625: I don't understand why this is not adding all tracked files, since I would think we don't want any duplicates.

robmen: Could do that but it would be hiding if `BuiltOutputs` were duplicated with inputs which would be very odd. This is the correct filter.

rseanhall: Sounds like you want a dictionary instead and generate an error in that scenario.

src/test/wix/TestData/WixprojPackageVcxprojWindowsApp/WixprojPackageVcxprojWindowsApp.wixproj
robmen: Interesting. The test did not fail. There is probably a warning in the test's build indicating that a higher version was selected. I'm going to try to make things a little better to handle the tests under the `build` folder.

src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
robmen: No. I just want the information in the tracking file if it ends up there.

rseanhall: So the answer is it's not perfect, but it works. Fine.

src/test/wix/TestData/WixprojPackageVcxprojWindowsApp/WixprojPackageVcxprojWindowsApp.wixproj
rseanhall: It succeeded on AppVeyor, too. So you could merge this now and fix it later if you want.

src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
robmen: To be clear, it would be internal code to catch if there was something wrong done in the backend earlier. It would throw an exception not display an error for the user to work with because it should never happen. It's the sort of thing that'd go behind `#if DEBUG`. Not worth the extra overhead, IMHO.

src/test/wix/TestData/WixprojPackageVcxprojWindowsApp/WixprojPackageVcxprojWindowsApp.wixproj
robmen: Lets' see if this last commit makes everything work consistently.

src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
rseanhall: To be clear, multiple rows in the File table can have the same input path. I still stand by original statement, I don't think we want duplicate Input tracked files.

robmen: The tracking file doesn't end up with duplicates. That's handled at the very end (during layout) when the complete list of tracked files is known. At this point in the code, we're not throwing away information that could be used (for example, an extension that wanted to report that multiple rows shared the same input path).

That's why this filters out only the intermediate and temporary files that also end up as input files.

rseanhall: You are presenting conflicting requirements.

1. `ITrackedFile`s don't have a 1:1 mapping with entries in the tracking file. Extensions should be informed of every time an input path is used. Under this requirement, no filtering should happen here and it should be done during layout instead.
2. An Input `ITrackedFile` should never be created if a Temporary or Intermediate `ITrackedFile` already exists with that input path. Under this requirement, it seems that `ITrackedFile`s are supposed to have a 1:1 mapping with entries in the tracking file. Which means throughout the code we should be preventing duplicates.

robmen: I'm not at either of those extremes. In this case here, my goal was for the WindowsInstaller backend to finish the binding of the database where Intermediates and Temporary tracked files are not also Inputs. That represents a reasonable "world-view" given the way this code treats its Intermediates and Temporary files.

Option 1 is definitely a reasonable world-view but it would require that everything down the pipeline know to not treat Intermediates and Temporary files from the WindowsInstaller Backend as Inputs. Now option 1 may very well become the best overall world-view after the Burn Backend is reviewed for incremental support. If so, then I'd want to revisit this decision here.

rseanhall: > but it would require that everything down the pipeline know to not treat Intermediates and Temporary files from the WindowsInstaller Backend as Inputs

What about parts of the pipeline that want to know when an Intermediate or Temporary file is used as an input? These are conflicting requirements, not extremes. Your objection might be overcome by adding a new type of tracked file, like IntermediateInput.

PR #127, opened 2022-02-26
Support certs on remote payloads and generate them from burn subcommand

robmen:
Bring back Authenticode certificate validation but only on Exe and
Msu remote payloads. Move the generation of remote payload XML to
a subcommand of the "burn command".

Fixes wixtoolset/issues#6447


robmen:
> Ideally the Netfx packages would be updated with the Authenticode signature information, which would get runtime testing on this through the Layout E2E tests.

That's a good idea.

rseanhall:
I guess we'll discuss this in the next meeting. I don't understand why the mutual exclusiveness part of the issue was reverted. And why the subcommand is creating source code that requires a 4.x feature to use.

rseanhall:
Something to watch out for: `WixToolsetTest.CoreIntegration.RemotePayloadFixture.CanGetRemotePayloadsRecursive` is always failing on Appveyor.

PR #132, opened 2022-03-02
Set ExePackage/@Uninstallable in bundle manifests

robmen:
Part of fix for wixtoolset/issues#6459


rseanhall (approved):
Yes, thanks. This still shouldn't close the issue unless you're going to update exepackage.cpp and add tests to PlanTest.cpp.

PR #137, opened 2022-03-08
Configure crash dumps and get them and pdbs in the logs artifact

rseanhall:


rseanhall:
This was built twice in my repo and twice here. It passed in all 4, and got a crash dump in 3 of them (both of the Actions builds and one Appveyor).

PR #140, opened 2022-03-14
Make engine skip planning if there are any downgrade related bundles

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/6677, https://github.com/wixtoolset/issues/issues/6722
Reverts https://github.com/wixtoolset/issues/issues/6537


src/burn/engine/plan.cpp
```diff
@@ -1266,18 +1312,72 @@ extern "C" HRESULT PlanDefaultRelatedBundleRequestState(
...
+        hr = PlanDefaultRelatedBundlePlanType(pRelatedBundle->detectRelationType, pRegistration->pVersion, pRelatedBundle->pVersion, &pRelatedBundle->defaultPlanRelationType);
+        ExitOnFailure(hr, "Failed to get default plan type for related bundle.");
+
+        pRelatedBundle->planRelationType = pRelatedBundle->defaultPlanRelationType;
+
+        hr = UserExperienceOnPlanRelatedBundleType(pUserExperience, pRelatedBundle->package.sczId, &pRelatedBundle->planRelationType);
+        ExitOnRootFailure(hr, "BA aborted plan related bundle type.");
+
+        if (BOOTSTRAPPER_RELATED_BUNDLE_PLAN_TYPE_DOWNGRADE == pRelatedBundle->planRelationType &&
+            pRelatedBundle->fPlannable && !fUninstalling && BOOTSTRAPPER_RELATION_UPGRADE != relationType)
```
rseanhall, line 1362: I wasn't sure which way to go here. If the related bundle isn't plannable (registered in ARP, but missing from the package cache so it can't be run), should it block the current bundle from downgrading it?

src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -1089,6 +1065,20 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
...
+        __in HRESULT /*hrRecommendation*/,
+        __in HRESULT* phrStatus
+        )
+    {
+        HRESULT hr = S_OK;
+
+        if (m_fSuppressDowngradeFailure)
+        {
+            *phrStatus = S_OK;
+        }
```
rseanhall, line 1078: I wonder if this should be a feature in the engine instead of wixstdba.

PR #143, opened 2022-03-15
Add WixBundleCommandLineAction

rseanhall:
Also fix themes for UnsafeInstall action.

Fixes https://github.com/wixtoolset/issues/issues/6736


robmen:
Ahh, good. I was afraid comments on a commit would not get promoted to the PR.

PR #145, opened 2022-03-16
Warn when an MSI or cabinet installed by an MSI is too large

robmen:
Fixes wixtoolset/issues#6408

### DIFF
--- a/src/wix/WixToolset.Core.Burn/Bundles/ProcessMsiPackageCommand.cs
+++ b/src/wix/WixToolset.Core.Burn/Bundles/ProcessMsiPackageCommand.cs
@@ -63,6 +63,8 @@ namespace WixToolset.Core.Burn.Bundles
             var compressed = false;
             try
             {
+                this.CheckIfWindowsInstallerFileTooLarge(packagePayload.SourceLineNumbers, sourcePath, "MSI");
+
                 using (var db = new Database(sourcePath, OpenDatabase.ReadOnly))
                 {
                     // Read data out of the msi database...
@@ -397,6 +399,8 @@ namespace WixToolset.Core.Burn.Bundles
                                     Packaging = packagePayload.Packaging,
                                     ParentPackagePayloadRef = packagePayload.Id.Id,
                                 });
+
+                                this.CheckIfWindowsInstallerFileTooLarge(this.Facade.PackageSymbol.SourceLineNumbers, payloadSourceFile, "cabinet");
                             }
                         }
                     }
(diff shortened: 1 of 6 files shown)


src/wix/test/WixToolsetTest.CoreIntegration/MsiFixture.cs
```diff
@@ -873,7 +873,7 @@ public void FailsBuildAtLinkTimeForMissingEnsureTable()
                     first =>
                     {
                         Assert.Equal(MessageLevel.Error, first.Level);
-                        Assert.Equal("The identifier 'WixCustomTable:TableDefinitionNotExposedByExtension' could not be found. Ensure you have typed the reference correctly and that all the necessary inputs are provided to the linker.", first.ToString());
+                        Assert.Equal("Cannot find the table definitions for the 'TableDefinitionNotExposedByExtension' table.  This is likely due to a typing error or missing extension.  Please ensure all the necessary extensions are supplied on the command line with the -ext parameter.", first.ToString());
```
rseanhall, line 876: Please change the name of the test to match the new expected behavior. `FailsBuildAtBindTimeForMissingEnsureTable`?

rseanhall (commented):
I would like to see a test that shows how to ensure an empty custom table that is defined in a Fragment instead of by an extension.

rseanhall:
Also, need to verify Bundle container .cabs are the right size.

src/wix/WixToolset.Core.Burn/Bundles/ProcessMsiPackageCommand.cs
```diff
@@ -397,6 +399,8 @@ private void ImportExternalCabinetAsPayloads(Database db, WixBundlePayloadSymbol
                                     Packaging = packagePayload.Packaging,
                                     ParentPackagePayloadRef = packagePayload.Id.Id,
                                 });
+
+                                this.CheckIfWindowsInstallerFileTooLarge(this.Facade.PackageSymbol.SourceLineNumbers, payloadSourceFile, "cabinet");
```
robmen, line 403: Doesn't this check if the cabinet files are too large in the bundle?

rseanhall: Yes, for `MsiPackage` files. But another check is needed for containers created by  https://github.com/wixtoolset/wix4/blob/73abf1db11a2ec38937682563d382a5c67fe2c40/src/wix/WixToolset.Core.Burn/Bundles/CreateContainerCommand.cs, right?

robmen: Not as long as Burn can handle the files. My understanding is that MSI can't handle cabinet files over 2GB not Burn.

rseanhall: I have not tested it. I guess it makes sense that if wixnative can create the cab then Burn should be able to read it, but I'm not 100% sure on that.

PR #153, opened 2022-03-30
Fix "flaky" test in DUtilUnitTest.

cpuwzd:


cpuwzd:
[Disassembly with WixAssert.txt](https://github.com/wixtoolset/wix4/files/8383381/Disassembly.with.WixAssert.txt)
[Disassembly with Xnit.Assert.txt](https://github.com/wixtoolset/wix4/files/8383382/Disassembly.with.Xnit.Assert.txt)

src/libs/dutil/test/DUtilUnitTest/DUtilTests.cpp
```diff
@@ -11,8 +11,8 @@ namespace DutilTests
     public ref class DUtil
     {
     public:
-        [Fact(Skip = "Flaky")]
-        void DUtilTraceErrorSourceFiltersOnTraceLevel()
+        [Fact]
```
rseanhall, line 14: I'm not an expert at assembly, but the information that you provided seems to confirm my theory that this won't change the behavior.

Xunit:

```
07FA7BCF  call        dword ptr [Pointer to: Xunit.Assert.Throws[[System.__Canon, mscorlib]](System.Action) (08069574h)]
```

WixAssert:

```
086D77DF  call        WixBuildTools.TestSupport.WixAssert.Throws[[System.__Canon, mscorlib]](System.Action) (086D7528h)
```

I'm also not an expert on the C++/CLI compiler so I can't explain how exactly it works. But if it compiles without error, then I trust that it didn't throw away code. Like I said on wix-devs, Visual Studio needs to support C++/CLI in order to provide all of its features like Intellisense and setting breakpoints. I don't want to stop skipping this test until the real issue causing it to fail is resolved.

src/libs/dutil/test/DUtilUnitTest/DUtilTests.cpp
```diff
@@ -21,7 +21,12 @@ namespace DutilTests
             Dutil_TraceSetLevel(REPORT_DEBUG, FALSE);
 
             Action^ action = gcnew Action(this, &DUtil::CallDutilTraceErrorSource);
-            Assert::Throws<Exception^>(action);
+            // xxxxx The following line is ambiguous when used with the current version of Xunit.
+            // xxxxx Xunit::Assert::Throws<Exception^>(action);
+            // xxxxx It has therefore been, at least temporarily, replaced by the next line of code.
+            // xxxxx See the comments in WixBuildTools.WixAssert for details.
```
rseanhall, line 27: We don't write comments like `// xxxxx {Text}` in WiX. Please remove the `xxxxx`s. Also, we would normally add a blank line above the comments and no blank line between the comment and its code but you have reversed that here.

src/internal/WixBuildTools.TestSupport/WixAssert.cs
```diff
@@ -91,5 +91,54 @@ public int GetHashCode(object obj)
...
+        /// without getting an ambiguous overload error.  If the specified test code
+        /// fails to generate an exception of the exact specified type, an assertion
+        /// exception is thrown. Otherwise, execution flow proceeds as normal.
+        /// </summary>
+        /// <typeparam name="T">The type name of the expected exception.</typeparam>
+        /// <param name="testCode">An Action delegate to run the test code.</param>
+        public static new void Throws<T>(System.Action testCode)
+            where T : System.Exception
+        {
+            Xunit.Assert.Throws<T>(testCode);
```
rseanhall, line 123: If you want to be explicit with `Xunit.Assert`, then I would move these into a separate class. That would also make it possible to not need the `new` keyword to avoid hiding the base class's version. I'm actually a little surprised that the C++/CLI can handle the `new` functionality of hiding the other one if it can't handle the overloading problem.

cpuwzd:
I believe that this pull request is ready to be merged. I updated the explanatory comments and followed Sean's guidance as best I could.

src/libs/dutil/test/DUtilUnitTest/DUtilTests.cpp
```diff
@@ -21,7 +21,13 @@ namespace DutilTests
...
 
             Action^ action = gcnew Action(this, &DUtil::CallDutilTraceErrorSource);
-            Assert::Throws<Exception^>(action);
+            // The following line is ambiguous when used with the current version of Xunit.
+            // Xunit::Assert::Throws<Exception^>(action);
+            // It has therefore been, at least temporarily, replaced by the next line of code.
+            // See the comments in WixBuildTools.WixAssert for details.
+            WixAssert::Throws<Exception^>(action);
+
+            WixBuildTools::TestSupport::WixAssert::Throws<Exception^>(action);
```
rseanhall, line 30: Please remove this line and change `Void` back to `void`. I can take this, but I fully expect to have to put it back to skipped.

PR #78, opened 2021-12-31
Various burn improvements

rseanhall:
https://github.com/wixtoolset/issues/issues/5796

PR #154, opened 2022-03-30
3693 -BundlePackage

rseanhall:
This is ready for review, though I will implement some of the TODOs before merging.

Closes https://github.com/wixtoolset/issues/issues/3693

Other PRs that contributed towards 3693:
      https://github.com/wixtoolset/wix4/pull/78
      https://github.com/wixtoolset/wix4/pull/85
      https://github.com/wixtoolset/wix4/pull/96
      https://github.com/wixtoolset/wix4/pull/128


src/burn/engine/package.cpp
```diff
@@ -371,7 +378,7 @@ extern "C" void PackageUninitialize(
     switch (pPackage->type)
     {
     case BURN_PACKAGE_TYPE_BUNDLE:
-        BundlePackageEnginePackageUninitialize(pPackage);
+        BundlePackageEnginePackageUninitialize(pPackage); // TODO: Modularization
```
robmen, line 381: What does `Modularization` mean here? I'm guessing it doesn't mean append the Merge Module GUID to the end of primary keys. :)

src/wix/WixToolset.Core.Burn/Bundles/ProcessBundlePackageCommand.cs
```diff
@@ -0,0 +1,152 @@
...
+                    case BurnCommon.IMAGE_FILE_MACHINE_ARM:
+                    case BurnCommon.IMAGE_FILE_MACHINE_ARMNT:
+                    case BurnCommon.IMAGE_FILE_MACHINE_I386:
+                        break;
+                    case BurnCommon.IMAGE_FILE_MACHINE_AMD64:
+                    case BurnCommon.IMAGE_FILE_MACHINE_ARM64:
+                        bundlePackage.Win64 = true;
+                        break;
+                    default:
+                        Debug.Assert(false, "Unknown machine type");
```
robmen, line 83: Since `Debug.Assert()` disappears in Release builds, I generally recommend throwing `InvalidOperationException` or `ArgumentException` or something with the error message so that we likely get a bug if/when it happens.

src/wix/WixToolset.Core/Compiler_Bundle.cs
```diff
@@ -2422,6 +2444,18 @@ private string ParseChainPackage(XElement node, WixBundlePackageType packageType
...
                 {
+                    case WixBundlePackageType.Bundle:
+                        WixBundleBundlePackageAttributes bundleAttributes = 0;
+
+                        this.Core.AddSymbol(new WixBundleBundlePackageSymbol(sourceLineNumbers, id)
+                        {
+                            Attributes = bundleAttributes,
+                            InstallCommand = installArguments,
+                            RepairCommand = repairArguments,
+                            UninstallCommand = uninstallArguments,
```
robmen, line 2455: Should we default the `UninstallCommand` for Bundles if they are not permanent?

src/burn/engine/package.cpp
rseanhall: It means whatever it means for the other existing package types. You tell me what that originally meant :)

robmen: Heh. Only guesses. 👻

src/wix/WixToolset.Core.Burn/Bundles/ProcessBundlePackageCommand.cs
rseanhall: I'm really not a fan of blocking people from building a bundle because one of the packages couldn't be harvested. It's looking like Burn's going to ignore this anyway until/unless it actually starts verifying that the bundle it found in the registry was the right bitness.

src/wix/WixToolset.Core/Compiler_Bundle.cs
rseanhall: The Burn engine automatically adds `/uninstall`. `BundlePackage`s are always uninstallable and always repairable.

robmen: Cool. Two questions:
1. Should `UninstallCommand` be additional arguments to `/uninstall` in the engine or if specified should it take over fully?
2. I can see scenarios where devs will want to not repair the nested bundle. Today with ExePackage that is done by not adding the `RepairArguments`. Is this supported for BundlePackage without custom BA logic?

rseanhall: `BundlePackage` is a lot closer to `MsiPackage` than `ExePackage` for these things.
1. Additional. Similar to `MsiProperty` for `MsiPackage` - the engine is going to add certain ones that the bundle author and BA have no control over.
2. There's the same problem with `MsiPackage`. It sounded like we were punting that to v5.

src/wix/WixToolset.Core.Burn/Bundles/ProcessBundlePackageCommand.cs
robmen: Maybe a warning then so the dev has some indication that harvesting didn't behave normally. I'm concerned about the future if this list changes and it silently slips through rather than noisly failing so we can find and fix it.

rseanhall: I guess. I don't imagine Windows or WiX supporting additional machine types anytime soon.

robmen: I would have said the same thing before ARM came along. :)  But I've been bitten by silent `Debug.Assert()`s enough that I don't care for them (unless they are there to provide useful developer context for a crash that is coming when violated).  :)

src/wix/WixToolset.Core/Compiler_Bundle.cs
robmen: 1. Cool. I wonder if we should name the attribute differently so developers don't set  `UninstallArguments="-uninstall"`  regularly.
2. True but redists are more likely to be shipped as Exes and Bundles than MSI packages. For example, `vc_redist` is a Bundle. The automatic detection of the Bundle will be awesome but most devs would not want to repair `vc_redist` with their bundle. I think `BundlePackage` should be considered more like `ExePackage` than `MsiPackage`.

rseanhall: Sounds like we need to design and implement `RepairCondition` into v4 then.

src/wix/WixToolset.Core.Burn/Bundles/ProcessBundlePackageCommand.cs
rseanhall: I mean, ARM/ARM64 has been a widely used architecture for over 10 years and there's still no stable WiX bundle for it. To be fair Windows hasn't supported for that long, but still.

src/wix/WixToolset.Core/Compiler_Bundle.cs
robmen: > Sounds like we need to design and implement RepairCondition into v4 then.

We could do that. But would the same repair (non)behavior as `ExePackage` work in this case? No `RepairArguments` means no repair.

rseanhall: No. The Burn engine knows how to do a repair, just like it knows how to repair an MsiPackage. We're not going to do weird things to avoid `RepairCondition`.

robmen: Okay.

However, I do think it is better to think of `BundlePackage` more like `ExePackage` than `MsiPackage`. The use cases for `BundlePackage` are much more aligned with `ExePackage`.

rseanhall: The use cases might be more aligned with `ExePackage`, but the capabilities of the package type itself is much more closely aligned with `MsiPackage`.

robmen: I understand that and I can see why you are approaching things the way you are given the Burn implementation. I'm approaching the developer experience side (use cases) and thus focused on how the language exposes all of the awesome Burn functionality.

rseanhall: I am also focused on making sure the language exposes all of the awesome Burn functionality, which is why I think it is a mistake to look towards `ExePackage` for inspiration when it is not a good example of how it can be used.

robmen: Okay, cool. 

I believe the most important use case to keep in mind is that most bundles contained in other bundles are redists. We've learned that repairing redists in the bundle is a bad idea. Repairing redists should be avoided enough that we might even consider making no-repair the default for bundles (though I'm not necessarily sold on that idea myself yet).

Less important but something I noodle on: bundles are always visible in ARP whereas MSI packages are invisible by default. That changes the interaction with uninstall since the user can manually execute it on nested bundles. At this point, I do think that the `Permanent` attribute plus dependency tracking exposes the right control for uninstall of nested bundles. Of course, if we hid nested bundles by default lots would need to be reconsidered.

rseanhall: For BundlePackage ARP visibility, that is a feature tracked in https://github.com/wixtoolset/issues/issues/4454. I was going to implement that after merging, so if you have thoughts on that please put them there.

robmen: Interesting. I was not expecting 4454 to be moved into `v4.0`.

Hiding nested bundles changes quite a few things. If hiding nested bundles was the default then `BundlePackage` _would_ be very much like `MsiPackage`. But I'm not sure nested bundles should be hidden by default. We'll definitely need to think through all the ramifications of hidden bundles. For example, I'm not sure nested bundles should be hidden by default.

robmen: Clearly... I'm not sure nested bundles should be hidden by default. 🤣

src/burn/engine/bundlepackageengine.cpp
```diff
@@ -92,45 +209,95 @@ extern "C" void BundlePackageEnginePackageUninitialize(
...
-    // update detect state
-    pPackage->currentState = fDetected ? BOOTSTRAPPER_PACKAGE_STATE_PRESENT : BOOTSTRAPPER_PACKAGE_STATE_ABSENT;
-
     if (pPackage->fCanAffectRegistration)
     {
         pPackage->installRegistrationState = BOOTSTRAPPER_PACKAGE_STATE_ABSENT < pPackage->currentState ? BURN_PACKAGE_REGISTRATION_STATE_PRESENT : BURN_PACKAGE_REGISTRATION_STATE_ABSENT;
     }
 
-    ReleaseRegKey(hkRegistration);
+    // TODO: The bundle is registering itself as a dependent when installed as a chain package, which prevents us from uninstalling it.
```
rseanhall, line 294: I believe that the `-parent` switch is what's supposed to prevent the bundle from registering itself as a dependent. Can someone confirm that this is the correct usage of that switch? Or is this going to need to be a new feature because that switch is for something else?

PR #155, opened 2022-03-31
Provide managed CA and Embedded UI DTF libraries via NuGet

robmen:
Plus a couple other small fixes and porting of fix from WiX v3.

Closes https://github.com/wixtoolset/issues/issues/6080
Fixes https://github.com/wixtoolset/issues/issues/4688


src/test/test.cmd
```diff
@@ -13,6 +13,8 @@
 
 @call burn\test_burn.cmd %_C% %_T% || exit /b
 
+msbuild -t:Restore dtf\DtfE2ETests.sln -p:Configuration=%_C% -nologo -m -warnaserror -bl:%_L%\dtfe2etests.binlog || exit /b
```
rseanhall, line 16: You just want to Restore, not build or test?

robmen: Heh, no. I'll change that in my next PR.

PR #158, opened 2022-04-13
Remote support for BundlePackage

rseanhall:
Fix discarding RollbackBoundary.
Only allow MsiPackage and MspPackage inside MSI transactions.


src/api/wix/WixToolset.Data/Symbols/WixBundleSymbol.cs
```diff
@@ -243,14 +250,52 @@ public WixBundleCommandLineVariables CommandLineVariables
...
         }
 
-        public PackagingType DefaultPackagingType => (this.Compressed.HasValue && !this.Compressed.Value) ? PackagingType.External : PackagingType.Embedded;
-
-        public bool DisableModify => (this.Attributes & WixBundleAttributes.DisableModify) == WixBundleAttributes.DisableModify;
+        public WixBundleModifyType DisableModify
+        {
+            get
+            {
+                if (Enum.TryParse((string)this.Fields[(int)WixBundleSymbolFields.DisableModify], true, out WixBundleModifyType value))
```
robmen, line 257: We've switched to using the numeric values for enums instead of the string values underlying the Data API. I believe there are a couple of other stragglers left in the Bundle symbols if you wanted to grab them as part of this change, that'd be awesome.

rseanhall:
I'm just following patterns. It doesn't seem to be worth my time discussing what the patterns should be given past experiences.

robmen:
I'm sorry you feel that way, but that is fine. I'll fix it later.

PR #175, opened 2022-06-02
LogVerifier/StyleCopComplaints

cpuwzd:
Attention Sean Hall:

LogVerifier.cs generates StyleCop errors when opened in Util.wixext.sln. These
changes appease StyleCop. Some errors can't be fixed by changing .editorconfig.
For example, requiring and disallowing "this." qualification are mutually 
exclusive and there appears to be no permissive option to allow both extremes. All of these changes are to LogVerifier.cs, itself and no changes have been made
 to .editorconfig .

### DIFF
--- a/src/test/burn/WixTestTools/LogVerifier.cs
+++ b/src/test/burn/WixTestTools/LogVerifier.cs
@@ -14,7 +14,7 @@ namespace WixTestTools
     public class LogVerifier
     {
         // Member Variables
-        private string logFile;
+        private readonly string logFile;
 
         /// <summary>
         /// Prevent creation of LogVerifier without log file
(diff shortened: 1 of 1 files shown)


cpuwzd:
I meant to reference MsiE2ETests.sln, rather than Util.wixext.sln. My bad.

rseanhall:
Thanks!

You should always treat .editorconfig as the source of truth. A lot of the settings in there are new. The codebase is massive so it's not practical to fix all the errors at once which leads to us incrementally fixing things as they come up like you did here.

rseanhall:
Thanks!

You should always treat .editorconfig as the source of truth. A lot of the settings in there are new. The codebase is massive so it's not practical to fix all the errors at once which leads to us incrementally fixing things as they come up like you did here.

PR #180, opened 2022-06-16
Add Github Actions workflow to upload a test report

rseanhall:
It has to be a separate workflow because the normal PR workflow doesn't have the required permissions to add the check for the test report. The test report will not work on this PR because triggered workflows like this only run when it is already in the default branch. Some examples can be seen at https://github.com/rseanhall/wix4/runs/6924360508 and https://github.com/rseanhall/wix4/runs/6924837814.

### DIFF
--- a/.github/workflows/test-report.yml
+++ b/.github/workflows/test-report.yml
@@ -1,0 +1,22 @@
+name: Build WiX Toolset v4 Test Report
+
+on:
+  workflow_run:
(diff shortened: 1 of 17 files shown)


robmen (approved):
Very nice. Was going to look into this at some point, glad you beat me to it.

rseanhall:
Thanks.

PR #184, opened 2022-06-29
Fix E2E test exception.

cpuwzd:
Upgrade xunit.runner.visualstudio from V2.4.1 to V2.4.5.


rseanhall:
I can take this, but I don't think it will fix the issue you were seeing. I should have already fixed that in my PR (https://github.com/wixtoolset/wix4/pull/182).

rseanhall:
Thanks!

PR #195, opened 2022-07-10
Support bind variables in Package and Bundle versions

robmen:
This was way more involved that I anticipated and demonstrates some of the more complex scenarios we'll deal with when bringing more compiler error handling to the backends. If I'd known the complexity before starting, I probably would've postponed the work to v5. But the work is instructive for the future and does fix up a few things, like Software Tag handling no longer dependent on properties.

Also, search for WiX Extensions with .wixext.dll and build DTF E2E tests

Closes wixtoolset/issues#6779


src/wix/WixToolset.Core/ExtensibilityServices/BackendHelper.cs
```diff
@@ -119,5 +119,40 @@ public bool IsValidShortFilename(string filename, bool allowWildcards)
...
+
+                if (strict || wixVersion.HasPatch)
+                {
+                    parsedVersion += $".{wixVersion.Patch}";
+                }
+
+                if (!strict && wixVersion.HasRevision)
+                {
+                    parsedVersion += $".{wixVersion.Revision}";
+                }
```
rseanhall, line 149: Am I reading this correctly? An MSI with `ProductVersion` v1.2.3.4 with default `MajorUpgrade` will get 1.2.3 in the upgrade table, which will make Burn think they're not the same.

robmen: You are reading that correctly. The Upgrade table versions only support 3-part versions, the fourth part is ignored so adding that part of the version is conceptually misleading. We could add additional data in the fourth field if it helps Burn implemention additional versioning features but there shouldn't be exact comparisons between `ProductVersion` and Upgrade table rows.

rseanhall: Burn v3 and v4 considers the fourth field as significant when figuring out whether the package is a downgrade or an upgrade. We have discussed changing Burn to only pay attention to the first three fields like Windows Installer but it didn't seem like anyone wanted that.

rseanhall: It considers the fourth field in both the ProductVersion and the Upgrade table rows.

robmen: Great. Then those versions should be WixVersion in Burn. I'll allow the full range ProductVersion in the Upgrade table so Burn can take advantage of it. Depending on the decision to #6788 then Burn may have lots of information to process them.

PR #197, opened 2022-07-14
Support remote Payloads

robmen:
Closes wixtoolset/issues#5601
Closes wixtoolset/issues#6802

### DIFF
--- a/src/wix/test/WixToolsetTest.CoreIntegration/PayloadFixture.cs
+++ b/src/wix/test/WixToolsetTest.CoreIntegration/PayloadFixture.cs
@@ -201,5 +201,86 @@ namespace WixToolsetTest.CoreIntegration
                 }, containers);
             }
         }
+
+        [Fact]
+        public void CanBuildBundleWithRemotePackagePaylod()
+        {
+            var folder = TestData.Get(@"TestData");
+
+            using (var fs = new DisposableFileSystem())
+            {
+                var baseFolder = fs.GetFolder();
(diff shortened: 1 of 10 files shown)


src/wix/test/WixToolsetTest.CoreIntegration/PayloadFixture.cs
```diff
@@ -201,5 +201,86 @@ public void ReplacesDownloadUrlPlaceholders()
...
+                    Path.Combine(folder, "SimpleBundle", "MultiFileBootstrapperApplication.wxs"),
+                    "-bindpath", Path.Combine(folder, "SimpleBundle", "data"),
+                    "-bindpath", Path.Combine(folder, ".Data"),
+                    "-intermediateFolder", intermediateFolder,
+                    "-o", bundlePath,
+                });
+
+                WixAssert.CompareLineByLine(new[]
+                {
+                    "The Payload 'RemotePayload' is being added to Container 'WixUXContainer', overriding its Compressed value of 'no'.",
```
rseanhall, line 280: On second thought, the linker should not be allowing remote payloads to be changed to compressed. This can happen for any container, not just WixUXContainer. I don't know whether you want to address that in this PR or create a separate issue. Other containers could skip assigning the payload to itself, which would be a new warning (LinkerWarnings.RemotePayloadInContainer).

robmen: I think that'd be best discussed in a separate issue.

PR #200, opened 2022-07-21
Add standard Variable name validation to util:Searches and SetVariable

rseanhall:
Ignore WixFragment symbol in GenerateManifestDataFromIRCommand.

Closes https://github.com/wixtoolset/issues/issues/6819


src/api/wix/WixToolset.Data/Symbols/WixBundleVariableSymbol.cs
```diff
@@ -35,6 +35,7 @@ public enum WixBundleVariableAttributes
         None = 0x0,
         Hidden = 0x1,
         Persisted = 0x2,
+        BuiltIn = 0x4,
```
robmen, line 38: Can variables not on the static builtin-list be declared as "builtin"? I can't think of a case where that makes sense, so I wonder if I'm missing a scenario.

src/wix/WixToolset.Core.Burn/Bundles/PerformBundleBackendValidationCommand.cs
```diff
@@ -173,7 +173,7 @@ private void ValidateSearch(WixSearchSymbol symbol)
 
         private void ValidateVariable(WixBundleVariableSymbol symbol)
         {
-            this.BackendHelper.ValidateBundleVariableName(symbol.SourceLineNumbers, "Variable", "Name", symbol.Id.Id);
+            this.BackendHelper.ValidateBundleVariableName(symbol.SourceLineNumbers, "Variable", "Name", symbol.Id.Id, allowBuiltIn: symbol.BuiltIn);
```
robmen, line 176: Is the only case the `allowBuiltIn` value is true is for the `BurnConstants.Xxx` values created in the compiler?

If so, perhaps those symbols should be created in the Burn backend instead of the compiler? From my scan of the diff, it seems `allowBuiltIn` bool to `ValidateionBundleVariableName` here and the `WixBundleVariableSymbol.BuiltIn` field could both be removed. But I'm still wondering if I'm missing a scenario for `WixBundleVariableSymbol.BuiltIn`.

rseanhall: Burn doesn't actually enforce the `Wix` prefix rule. I believe this is a reasonable implementation that helps the Burn part of the world. It hasn't received near as much attention as other parts so there are lots of places for improvements.

PR #205, opened 2022-07-29
Add two useful tools for debugging custom actions

cpuwzd:
This pull request includes two tools I have developed to aid in debugging
custom actions. The first tool provides a way to match up writes to and reads
from custom action data. The second tool is a simple, "Here I am!" that can be
pasted freely into custom action code to allow the analysis of code flow. It is especially useful when used inn conjunction with calls to WcaLog() to provide
data state in conjunction with the flow. There are a few comments in WcaUtil.h
to help with setting things up.


cpuwzd:
Why is the required build not running?

robmen:
I'm not sure these two changes add a lot of functionality beyond what is available by the Windows Installer. The `CHECKIN` feature doesn't add enough context to be committed to the source code. Thus they could be temporary `WcaLog()` additions that get removed before committing.

The `ANNOTATE` feature is a bit more interesting. The Windows Installer already displays the CustomActionData in the log file (when you remove the secure custom action attribute) so you can see what the CA code output. That makes it pretty straightforward to find the mismatch between writes and reads. However, I could see additional logging that makes it even easier to match up the writes and reads. But I would recommend designing that to always be on in `DEBUG` builds instead of requiring adding a custom define attribute that could be accidentally left on in Release builds and potentially leak secrets.

Finally, builds require approval by a project maintainer to ensure that bad actors don't send PRs that contain bitcoin mining changes in them and drive up GitHub's hosting bill for ill gains.

PR #209, opened 2022-08-03
Update expected message in CanLoadFDDx86EarliestCoreMBA test

rseanhall:
Fixes Appveyor build and probably Bob's, too.


barnson:
Can confirm. I still have the `The ordinal 345 could not be located...` error to investigate but that's the last problem I'm working around by not running tests...

PR #210, opened 2022-08-03
6836 - Make the clean room process replace the splash screen with its own

rseanhall:
This is required to get Windows to automatically transfer foreground focus so that the BA's window or the elevation prompt is automatically activated.

Fixes https://github.com/wixtoolset/issues/issues/6836


barnson:
That's not as bad as I was expecting...

rseanhall:
Same here. I was expecting to have to manually transfer the focus.

PR #211, opened 2022-08-03
Handle suppressed Options button.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/5624.

### DIFF
--- a/src/ext/Bal/wixstdba/Resources/HyperlinkLargeTheme.xml
+++ b/src/ext/Bal/wixstdba/Resources/HyperlinkLargeTheme.xml
@@ -22,7 +22,8 @@
         </Page>
         <Page Name="Install">
             <Label X="11" Y="80" Width="-11" Height="30" FontId="2" DisablePrefix="yes">#(loc.InstallHeader)</Label>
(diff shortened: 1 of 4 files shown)


src/ext/Bal/wixstdba/Resources/HyperlinkLargeTheme.xml
```diff
@@ -22,7 +22,8 @@
         </Page>
         <Page Name="Install">
             <Label X="11" Y="80" Width="-11" Height="30" FontId="2" DisablePrefix="yes">#(loc.InstallHeader)</Label>
-            <Label X="11" Y="121" Width="-11" Height="-129" FontId="3" DisablePrefix="yes">#(loc.InstallMessage)</Label>
+            <Label X="11" Y="121" Width="-11" Height="-129" FontId="3" VisibleCondition="WixStdBASuppressOptionsUI" DisablePrefix="yes">#(loc.InstallMessage)</Label>
+            <Label X="11" Y="121" Width="-11" Height="-129" FontId="3" VisibleCondition="NOT WixStdBASuppressOptionsUI" DisablePrefix="yes">#(loc.InstallMessageOptions)</Label>
```
rseanhall, line 26: I think it would be better if it has two child `Text` elements instead of duplicating the whole `Label`. This makes it easier for someone to customize an attribute without worrying about forgetting to change it on both.

PR #216, opened 2022-08-06
Create setup segment and WixAdditionalTools bundle

robmen:
The setup segment is scheduled after building the shipping WiX files to use and include the current build's tooling. The WixAdditionalTools bundle currently only contains the ThmViewer package but may include additional tools in the future.

At this time, the primary purpose is to produce a shipping MSI and bundle from the WiX v4 build process and dogfood the rest of the tooling.

Closes wixtoolset/issues#4919

Also, remove ProjectTypeGuids and use DefaultProjectTypeGuid instead.


src/setup/ThmViewerPackage/ThmViewerPackage.wxs
```diff
@@ -2,7 +2,7 @@
 <Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
   <Package Name="WiX Toolset Theme Viewer" Manufacturer="WiX Toolset" Language="1033" Version="!(bind.fileVersion.ThmViewerFile)" UpgradeCode="59c4b122-5167-445b-8fc4-09dcd4eced89">
     <MajorUpgrade DowngradeErrorMessage="A newer version of [ProductName] is already installed." />
-    <MediaTemplate EmbedCab="yes" />
+    <MediaTemplate />
```
barnson, line 5: Can be omitted entirely.

robmen: Right. Will fix next PR.

src/setup/WixAdditionalTools/WixAdditionalTools.wxs
```diff
@@ -0,0 +1,27 @@
...
+          AboutUrl="!(loc.AboutUrl)" HelpUrl="!(loc.SupportUrl)" UpdateUrl="!(loc.UpdateUrl)">
+    <BootstrapperApplication>
+      <bal:WixStandardBootstrapperApplication
+        LicenseUrl="!(loc.LicenseUrl)"
+        Theme="hyperlinkLicense"
+        ShowVersion="true"
+        LogoFile="wix-logo-sq.png" />
+    </BootstrapperApplication>
+
+    <SetVariable Variable="InstallFolder" Value="[ProgramFilesFolder]WiX Toolset v$(SetupMajorMinorVersion)\" />
```
rseanhall, line 15: Shouldn't we be persisting this variable?

robmen: Could do. But Windows Installer won't let us change the install location, so it doesn't have any effect. But it would probably look better in subsequent operation's log files if the value was the same.

src/setup/WixAdditionalTools/WixAdditionalTools.wxs
```diff
@@ -0,0 +1,27 @@
...
+    <!-- TODO: bring back SoftwareTag when #6854 fixed -->
+    <!-- <SoftwareTag Regid="!(loc.Regid)" InstallPath="[InstallFolder]" /> -->
+
+    <Update Location="!(loc.UpdateUrl)" />
```
rseanhall, line 19: FYI, wixstdba has no support today for this feature so this won't do anything.

PR #219, opened 2022-08-09
Generate metadata needed to populate update feed

robmen:
The first step towards wixtoolset/issues#5367


src/setup/MetadataTask/GenerateMetadata.cs
```diff
@@ -0,0 +1,113 @@
...
+            File.WriteAllText(metadataFilePath, json);
+        }
+
+        private static ArchitectureType PlatformToArchitecture(Platform platform)
+        {
+            switch (platform)
+            {
+                case Platform.X86: return ArchitectureType.X86;
+                case Platform.X64: return ArchitectureType.X86;
+                case Platform.ARM64: return ArchitectureType.X86;
```
rseanhall, line 108: Are they all supposed to return `ArchitectureType.X86`?

robmen: Doh! No. Will fix.

PR #239, opened 2022-08-29
Restrict panels to static controls.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/6635.


rseanhall:
Thanks. The intent of `BillboardPanel` was to allow schema validation that `X` and `Y` were not supposed to be specified on them. It makes me sad that this will be lost. I kind of want to make it an error when any of the control attributes are specified on `Billboard/Panel` elements.

barnson:
We can do context-sensitive validation without having a specific element name. https://github.com/wixtoolset/issues/issues/6893

PR #240, opened 2022-08-31
Supply hashes to BA if present in update feed.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/6353.

### DIFF
--- a/src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
+++ b/src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
@@ -106,6 +106,8 @@ namespace WixToolset.Mba.Core
         /// <param name="wzUpdateLocation"></param>
         /// <param name="dw64Size"></param>
         /// <param name="wzVersion"></param>
+        /// <param name="wzHash"></param>
+        /// <param name="hashAlgorithm"></param>
         /// <param name="wzTitle"></param>
         /// <param name="wzSummary"></param>
         /// <param name="wzContentType"></param>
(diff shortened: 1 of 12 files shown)


src/burn/engine/detect.cpp
```diff
@@ -397,11 +398,25 @@ static HRESULT DetectAtomFeedUpdate(
         for (DWORD i = 0; i < pApupChain->cEntries; ++i)
         {
             APPLICATION_UPDATE_ENTRY* pAppUpdateEntry = &pApupChain->rgEntries[i];
+            APPLICATION_UPDATE_ENCLOSURE* pEnclosure = pAppUpdateEntry->rgEnclosures;
 
-            hr = UserExperienceOnDetectUpdate(pUX, pAppUpdateEntry->rgEnclosures ? pAppUpdateEntry->rgEnclosures->wzUrl : NULL, 
-                pAppUpdateEntry->rgEnclosures ? pAppUpdateEntry->rgEnclosures->dw64Size : 0, 
-                pAppUpdateEntry->pVersion, pAppUpdateEntry->wzTitle,
-                pAppUpdateEntry->wzSummary, pAppUpdateEntry->wzContentType, pAppUpdateEntry->wzContent, &fStopProcessingUpdates);
+            if (pEnclosure)
```
rseanhall, line 403: Shouldn't it check whether there's a hash and/or the value of `digestAlgorithm`?

src/api/burn/WixToolset.Mba.Core/IBootstrapperApplication.cs
```diff
@@ -2150,4 +2154,21 @@ public enum INSTALLUILEVEL
         /// </summary>
         UacOnly = 0x200,
     }
+
+    /// <summary>
+    /// The available types of hashes.
+    /// </summary>
+    public enum BOOTSTRAPPER_UPDATE_HASH_TYPE
```
rseanhall, line 2161: This enum already exists as `UpdateHashType` in IBootstrapperEngine.cs which already existed in wix3. It looks like I missed updating that enum when changing to SHA512.

src/burn/engine/detect.cpp
barnson: The ternary covers the algorithm but I coerced a null to an empty.

PR #244, opened 2022-09-10
WixStdBA bundle updates

barnson:


src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -342,6 +348,78 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
...
+
+            hr = BalSetVersionVariable(WIXSTDBA_VARIABLE_UPDATE_AVAILABLE, wzUpdateVersion);
+            BalExitOnFailure(hr, "Failed to set WixStdBAUpdateAvailable value: %ls.", wzUpdateVersion);
+
+            hr = m_pEngine->SetUpdate(NULL, wzUpdateLocation, dw64Size, hashAlgorithm, wzHash);
+            BalExitOnFailure(hr, "Failed to set update location: %ls.", wzUpdateLocation);
+        }
+
+    LExit:
+        return __super::OnDetectUpdate(wzUpdateLocation, dw64Size, wzHash, hashAlgorithm, wzUpdateVersion, wzTitle, wzSummary, wzContentType, wzContent, pfCancel, pfStopProcessingUpdates);
```
rseanhall, line 419: This is throwing away the HRESULT from above. Do we want to do that?

src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -4786,6 +4884,27 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
         m_pControlUninstallButton = NULL;
         ++pAssignControl;
 
+        pAssignControl->wId = WIXSTDBA_CONTROL_CHECKING_FOR_UPDATES_LABEL;
+        pAssignControl->wzName = L"CheckingForUpdatesLabel";
+        pAssignControl->ppControl = &m_pControlCheckingForUpdatesLabel;
+        pAssignControl->dwAutomaticBehaviorType = dwAutomaticBehaviorType;
```
rseanhall, line 4890: This is conflicting with using the `EnableCondition` attribute on the label in the theme. If wixstdba owns whether it should be shown or not, then the `EnableCondition` shouldn't be there (it won't do anything) and the code above should use `m_pControlCheckingForUpdatesLabel` instead of getting the control through `ThemeControlExistsById`. If the user gets to decide, then this shouldn't be a known control to wixstdba.

rseanhall (commented):
I really would like to see E2E tests for this and the hash feature you implemented earlier. I guess tests for this PR would have to be manual, though. The UpdateBundleTests already start up a webserver that serves a hardcoded feed.

src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
barnson: This label doesn't have any conditions.

src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
```diff
@@ -342,6 +348,78 @@ class CWixStandardBootstrapperApplication : public CBalBaseBootstrapperApplicati
     }
 
 
+    virtual STDMETHODIMP OnDetectUpdateBegin(
+        __in_z LPCWSTR wzUpdateLocation,
+        __inout BOOL* pfCancel,
+        __inout BOOL* pfSkip
+    )
```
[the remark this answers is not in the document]
barnson: If we can't make .editorconfig follow the style, we should call for a Style Convention to change it. Not being able to do ^KD blows.

src/ext/Bal/wixstdba/WixStandardBootstrapperApplication.cpp
rseanhall: Sorry, I got it mixed up with ones below but it looks like you figured it out. I was trying to provide a variable in these situations so the user could use a condition instead of making wixstdba responsible for its visibility. That allows the user to provide multiple controls that are conditioned off of it. If we're going to keep it this way, then the code above needs to always change the visibility of the control in case the user doesn't have the default `Visible=no`.

barnson: The variable's set during Detect, so I didn't want to mess with refreshing the Loading page.

PR #252, opened 2022-09-23
Resolve Issue5371

cpuwzd:
This change makes it possible to add, modify and remove comments on user accounts
while installing a package.
Remove debugging tools


src/ext/Util/ca/scaexec.cpp
```diff
@@ -680,6 +707,34 @@ static HRESULT RemoveUserInternal(
...
+    {
+        er = ::DsGetDcNameW(NULL, (LPCWSTR)pwzDomain, NULL, NULL, NULL, &pDomainControllerInfo);
+        if (RPC_S_SERVER_UNAVAILABLE == er)
+        {
+            // MSDN says, if we get the above error code, try again with the "DS_FORCE_REDISCOVERY" flag
+            er = ::DsGetDcNameW(NULL, (LPCWSTR)pwzDomain, NULL, NULL, DS_FORCE_REDISCOVERY, &pDomainControllerInfo);
+        }
+        if (ERROR_SUCCESS == er)
+        {
+            *ppwzServerName = pDomainControllerInfo->DomainControllerName + 2;  //Add 2 so that we don't get the \\ prefix
```
robmen, line 725: Should we verify that the domain `DomainControllerName` actually starts with `\\` before removing the first two characters? The documentation says the name is prefixes with `\` but I'm guessing that is a translation hiccup and they really meant `\\`. Still, it'd be bad to accidentally access dead memory here if the domain name was a single letter and missing the `\\`.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +805,133 @@ extern "C" UINT __stdcall CreateUser(
...
+
+                            hr = WcaCaScriptWriteNumber(hRollbackScript, iRollbackAttributes);
+                            ExitOnFailure(hr, "Failed to add rollback attributes to rollback script.");
+
+                            // Nudge the system to get all our rollback data written to disk.
+                            WcaCaScriptFlush(hRollbackScript);
+                        }
+                    }
+
+				    er = SetUserPassword(pwzServerName, pwzName, pwzPassword);
```
robmen, line 894: Indentation needs to be fixed here. Maybe tabs were accidentally inserted here?

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +805,133 @@ extern "C" UINT __stdcall CreateUser(
...
+
+				    if (SCAU_REMOVE_COMMENT & iAttributes)
+				    {
+					    er = SetUserComment(pwzServerName, pwzName, L"");
```
robmen, line 898: The error code `er` needs to be validated or not set at all if the code intends to ignore failures (because it is all best effort).

src/ext/Util/ca/scaexec.cpp
```diff
@@ -520,55 +519,83 @@ static HRESULT ModifyUserLocalBatchRight(
...
+    {
+        *ppComment = L"";
+    }
+    else if (pwzComment && *pwzComment)
+    {
+        *ppComment = pwzComment;
+    }
+}
+
+static NET_API_STATUS SetUserPassword(__in LPWSTR pwzServerName, __in LPWSTR pwzName, __in LPWSTR pwzPassword)
```
robmen, line 573: It's beter that all functions return `HRESULT`.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -520,55 +519,83 @@ static HRESULT ModifyUserLocalBatchRight(
...
+{
+    DWORD dw;
```
robmen, line 575: This isn't used, so you can pass `NULL` to `NetUserSetInfo()` and avoid creating it.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +805,133 @@ extern "C" UINT __stdcall CreateUser(
...
+                            {
+                                WcaLogError(hr, "failed to get existing user rights: %ls, continuing anyway.", pwzName);
+                            }
+                            else
+                            {
+                                if (!(SCAU_ALLOW_LOGON_AS_SERVICE & iOriginalAttributes) && (SCAU_ALLOW_LOGON_AS_SERVICE & iAttributes))
+                                {
+                                    iRollbackAttributes |= SCAU_ALLOW_LOGON_AS_SERVICE;
+                                }
+                                if (!(SCAU_ALLOW_LOGON_AS_BATCH & iOriginalAttributes) && (SCAU_ALLOW_LOGON_AS_BATCH & iAttributes))
```
robmen, line 872: Nit, a blank line between unrelated `if`s helps show they aren't `else if`.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +805,133 @@ extern "C" UINT __stdcall CreateUser(
...
-                er = ::NetUserGetInfo(wz, pwzName, 1, reinterpret_cast<LPBYTE*>(&puserInfo));
+                pUserInfo1 = NULL;
+                er = ::NetUserGetInfo(pwzServerName, pwzName, 1, reinterpret_cast<LPBYTE*>(&pUserInfo1));
                 if (NERR_Success == er)
                 {
-                    // Change the existing user's password and attributes again then try
-                    // to update user with this new data
-                    SetUserPasswordAndAttributes(puserInfo, pwzPassword, iAttributes);
+                    // There is no rollback scheduled if the key is empty.
+			        // Best effort to get original configuration and save it in the script so rollback can restore it.
```
robmen, line 843: Indented too far.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -851,39 +947,37 @@ extern "C" UINT __stdcall CreateUser(
...
     }
 
-    //
-    // Add the users to groups
-    //
-    while (S_OK == (hr = WcaReadStringFromCaData(&pwz, &pwzGroup)))
-    {
-        hr = WcaReadStringFromCaData(&pwz, &pwzGroupDomain);
-        ExitOnFailure(hr, "failed to get domain for group: %ls", pwzGroup);
+//
```
robmen, line 950: This indentation should not be removed.

robmen (changes requested):
Pretty good, mostly nits to address. You don't have to worry about rebasing here since these changes are isolated.

robmen:
Could you amend the git commit message to be more like:

> Support add, modify and remove comments on user accounts
>
> Fixes 5371

Then in the PR use the actual reference to the bug number like so:

> Support add, modify and remove comments on user accounts
>
> Fixes wixtoolset/issues<ReplaceAngleBracketsAndEverythingInBetweenWithASingle#Character>5371

Notice: I didn't add the # character to prevent premature linking of this PR to the issue. Doing the above will create a clean git message that will automatically close the related issue when the PR is resolved.

barnson:
(You can add a URL to the issue; GitHub does the right thing.)

cpuwzd:
Bob,

I'm sorry,  but I don't know what URL you are talking about, how to add 
a URL to the issue, or what purpose
it would serve. A couple of sentences and possibly a concrete example 
would probably give me a better frame
of reference To understand your suggestion. Our backgrounds are vastly 
different and occasionally this matters.

Thanks for your efforts to improve my work.

Ron

barnson:
Rob was suggesting how you could link this PR with the underlying bug. The easiest way is to specify the URL to the issue: https://github.com/wixtoolset/issues/issues/5371. GitHub notices the link and will note that it was mentioned.

robmen:
GitHub documentation about that feature can be found here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +820,136 @@ extern "C" UINT __stdcall CreateUser(
...
+
+                        hr = WcaCaScriptWriteNumber(hRollbackScript, iRollbackAttributes);
+                        ExitOnFailure(hr, "Failed to add rollback attributes to rollback script.");
+
+                        // Nudge the system to get all our rollback data written to disk.
+                        WcaCaScriptFlush(hRollbackScript);
+                    }
+                }
+
+                ExitOnNetworkError(::SetUserPassword(pwzServerName, pwzName, pwzPassword));
```
robmen, line 911: If these are best effort, is it better to at least try the other `SetXxx` functions even if one fails? Also, it would probably be a good idea to log when the user info is not updated, especially if it also says", continuing anyway.".

src/ext/Util/ca/scaexec.cpp
```diff
@@ -747,95 +820,136 @@ extern "C" UINT __stdcall CreateUser(
...
+                    ExitOnNetworkError(NERR_UserExists);
+                }
             }
         }
         else if (NERR_PasswordTooShort == er || NERR_PasswordTooLong == er)
         {
             MessageExitOnFailure(hr = HRESULT_FROM_WIN32(er), msierrUSRFailedUserCreatePswd, "failed to create user: %ls due to invalid password.", pwzName);
         }
+
+    LNExit:
```
robmen, line 952: If you really need a jump, then pull out the necessary code to a separate function and use `LExit` normally. However, I _think_ it makes sense to try updating as much as possible so this extra label might not be necessary. As noted above, definitely don't want multiple jump points in a function.

robmen (changes requested):
Multiple jump points in a function are a no go.

It also looks like this change failed one of the tests.

robmen:
Here is a quick link to the failed test: https://github.com/wixtoolset/wix4/actions/runs/3120088311/jobs/5060515422#step:4:3093

src/ext/Util/ca/scaexec.cpp
cpuwzd: Done.

src/ext/Util/ca/scaexec.cpp
cpuwzd: Removed all tabs from the  file and replaced them with spaces.

src/ext/Util/ca/scaexec.cpp
cpuwzd: Done.

src/ext/Util/ca/scaexec.cpp
cpuwzd: Done.

src/ext/Util/ca/scaexec.cpp
cpuwzd: Point taken. Done.

src/ext/Util/ca/scaexec.cpp
cpuwzd: I think I fixed this, but I'm not sure I'm looking at the same indentation problem that you are.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include "precomp.h"
 
+// Exit macros
+#define ExitOnNetworkError(x) { er = x; if (ERROR_SUCCESS != er) { goto LNExit; }}
```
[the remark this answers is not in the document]
cpuwzd: Fixed. I find that the parallel use of er and hr makes the code harder to follow. Do you have any suggestions for resolving
er errors into hr errors as they occur?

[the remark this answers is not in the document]
cpuwzd: Would it be good, in general to use
hr = HRESULT_FROM_WIN32(er);
rather than letting the er error hang over us like a cloud?

src/ext/Util/ca/scaexec.cpp
cpuwzd: I've taken a shot at this. I may have erred on the side of leniency. Make sure that I did not allow the process to continue
when it should not.

src/ext/Util/ca/scaexec.cpp
cpuwzd: No jump needed.

robmen:
@cpuwzd this PR is failing to pass the tests. How much time do you want me to spend reviewing before getting the tests working again?

cpuwzd:
Rob,

I've been working on this situation for two weeks. The failing tests are 
among the MsiE2E tests, which do not run when running
devbuild. I have all but one of these tests running currently and I 
think I have solved the last one. There is no sense reviewing
the visible code until I do another push. It might be today.

Thanks for your patience.

Ron

robmen:
Good to hear you are still making progress. Diagnosing failures in runtime tests is one of the most complex dev tasks in WiX.

cpuwzd:
Looks good to go.

cpuwzd:
Rob,

I rebased this branch on wixtoolset/develop. Are there any outstanding issues?

PR #255, opened 2022-09-28
Fix UICompiler custom actions duplication

StefanStojanovic:
For each ui:WixUI tag found, UICompiler adds PrintEula and ValidatePath custom actions. This causes problems with duplicate symbols if 2 or more such tags are used. This change makes sure that UI custom actions are added only once in total.


StefanStojanovic:

PR #260, opened 2022-10-06
Use HTTPS instead of HTTP

mcraiha:
Fix some of these


mcraiha:

robmen:
Rather than just add `https`, it'd be great if you updated the links to the current pages. Otherwise, this doesn't really remove any redirects and doesn't help.

mcraiha:
I tried to avoid to update the links, since e.g. https://msdn.microsoft.com/library/aa372071.aspx has no Wayback Machine's archived page so I am not 100% sure that the new links contain same content. But I did it now.

PR #262, opened 2022-10-14
Add NetFx .NET compatibility check for MSI

StefanStojanovic:
Adds new custom element in NetFx extension for running NetCoreCheck.exe tool from within the MSI installer - `<netfx:DotNetCompatibilityCheck />`. The checks are run before evaluating launch conditions, so their results can be used in those conditions. There is no limitation on the number of checks that can be run, so installer may query various runtimes on different platforms and versions and with different roll forward policies.

Fixes https://github.com/wixtoolset/issues/issues/6264

### DIFF
--- a/src/ext/NetFx/ca/netfxca.cpp
+++ b/src/ext/NetFx/ca/netfxca.cpp
@@ -37,8 +37,8 @@ LPCWSTR vcsNgenStrongName =
 enum eNgenStrongName { ngsnName = 1, ngsnValue };
 
 LPCWSTR vscDotNetCompatibilityCheckQuery =
-    L"SELECT `Platform`, `RuntimeType`, `Version`, `RollForward`, `Variable` FROM `Wix4NetFxDotNetCompatibilityCheck`";
-enum eDotNetCompatibilityCheckQuery { platform = 1, runtimeType, version, rollForward, variable };
+    L"SELECT `Platform`, `RuntimeType`, `Version`, `RollForward`, `Property` FROM `Wix4NetFxDotNetCheck`";
+enum eDotNetCompatibilityCheckQuery { platform = 1, runtimeType, version, rollForward, property };
 
 // Searches subdirectories of the given path for the highest version of ngen.exe available
 static HRESULT GetNgenVersion(
@@ -454,30 +454,6 @@ LExit:
 }
 
 /******************************************************************
- GetNetCoreCheckBinaryId - gets NetCoreCheck key from Binary table
-
-********************************************************************/
-static HRESULT GetNetCoreCheckBinaryId(
-    __in LPCWSTR pwzPlatform,
-    __out LPWSTR* ppwzBinaryId
-    )
-{
-    HRESULT hr = S_OK;
-
-    if (0 == lstrcmpW(L"x86", pwzPlatform)) {
-        hr = StrAllocString(ppwzBinaryId, L"Wix4NetCheck_x86", 0);
-    } else if (0 == lstrcmpW(L"x64", pwzPlatform)) {
-        hr = StrAllocString(ppwzBinaryId, L"Wix4NetCheck_x64", 0);
-    } else if (0 == lstrcmpW(L"arm64", pwzPlatform)) {
-        hr = StrAllocString(ppwzBinaryId, L"Wix4NetCheck_arm64", 0);
-    } else {
-        hr = E_INVALIDARG;
-    }
-
-    return hr;
-}
-
-/******************************************************************
  SchedNetFx - entry point for NetFx Custom Action
 
 ********************************************************************/
@@ -527,6 +503,15 @@ extern "C" UINT __stdcall SchedNetFx(
     // initialize
     hr = WcaInitialize(hInstall, "SchedNetFx");
     ExitOnFailure(hr, "failed to initialize");
+
+    // If Wix4NetFxNativeImage table doesn't exist skip the rest of this custom action
+    hr = WcaTableExists(L"Wix4NetFxNativeImage");
+    if (S_FALSE == hr)
+    {
+        hr = S_OK;
+        ExitFunction();
+    }
+    ExitOnFailure(hr, "failed to check if table Wix4NetFxNativeImage exists");
 
     hr = GetNgenPath(&pwz32Ngen, FALSE);
     f32NgenExeExists = SUCCEEDED(hr);
@@ -920,78 +905,74 @@ extern "C" UINT __stdcall DotNetCompatibilityCheck(
     HRESULT hr = S_OK;
     UINT er = ERROR_SUCCESS;
 
-    DWORD dwTempPathLength = 0;
-    LPWSTR pwzTempPath = NULL;
     PMSIHANDLE hView = NULL;
     PMSIHANDLE hRec = NULL;
     LPWSTR pwzPlatform = NULL;
     LPWSTR pwzNetCoreCheckBinaryId = NULL;
+    LPWSTR pwzNetCoreCheckDirectoryName = NULL;
+    LPWSTR pwzNetCoreCheckDirectoryPath = NULL;
     LPWSTR pwzNetCoreCheckFilePath = NULL;
     LPWSTR pwzRuntimeType = NULL;
     LPWSTR pwzVersion = NULL;
     LPWSTR pwzRollForward = NULL;
-    LPWSTR pwzVariable = NULL;
+    LPWSTR pwzProperty = NULL;
     LPWSTR pwzCommandLine = NULL;
+    HANDLE hProcess = NULL;
+    DWORD dwExitCode = 0;
 
     // Initialize
     hr = WcaInitialize(hInstall, "DotNetCompatibilityCheck");
     ExitOnFailure(hr, "failed to initialize");
 
-    // Get temp directory path
-    dwTempPathLength = GetTempPathW(dwTempPathLength, NULL);
-    if (0 == dwTempPathLength)
+    // If Wix4NetFxDotNetCheck table doesn't exist skip the rest of this custom action
+    hr = WcaTableExists(L"Wix4NetFxDotNetCheck");
+    if (S_FALSE == hr)
     {
-        hr = HRESULT_FROM_WIN32(GetLastError());
-        ExitOnFailure(hr, "failed to get temp path length");
+        hr = S_OK;
+        ExitFunction();
     }
-
-    hr = StrAlloc(&pwzTempPath, ++dwTempPathLength);
-    ExitOnFailure(hr, "failed to allocate temp path variable");
-
-    dwTempPathLength = GetTempPathW(dwTempPathLength, pwzTempPath);
-    if (0 == dwTempPathLength)
-    {
-        hr = HRESULT_FROM_WIN32(GetLastError());
-        ExitOnFailure(hr, "failed to get temp path value");
-    }
+    ExitOnFailure(hr, "failed to check if table Wix4NetFxDotNetCheck exists");
 
     // Open view on .NET compatibility check table
     hr = WcaOpenExecuteView(vscDotNetCompatibilityCheckQuery, &hView);
-    ExitOnFailure(hr, "failed to open view on Wix4NetFxDotNetCompatibilityCheck table");
+    ExitOnFailure(hr, "failed to open view on Wix4NetFxDotNetCheck table");
 
     // Go through all records and run NetCorCheck.exe for each
     while (S_OK == (hr = WcaFetchRecord(hView, &hRec)))
     {
-        HANDLE hProcess = INVALID_HANDLE_VALUE;
-        DWORD dwExitCode = 0;
-
         // Extract NetCoreCheck.exe for platform to temp directory
         hr = WcaGetRecordString(hRec, platform, &pwzPlatform);
-        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Platform");
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.Platform");
 
-        hr = GetNetCoreCheckBinaryId(pwzPlatform, &pwzNetCoreCheckBinaryId);
+        hr = StrAllocFormatted(&pwzNetCoreCheckBinaryId, L"Wix4NetCheck_%ls", pwzPlatform);
         ExitOnFailure(hr, "failed to get NetCoreCheck binary id for platform %ls", pwzPlatform);
+        
+        hr = GuidCreate(&pwzNetCoreCheckDirectoryName);
+        ExitOnFailure(hr, "failed to set NetCoreCheck directory name");
+        
+        hr = PathCreateTempDirectory(NULL, pwzNetCoreCheckDirectoryName, 1, &pwzNetCoreCheckDirectoryPath);
+        ExitOnFailure(hr, "failed to make NetCoreCheck directory path for name %ls", pwzNetCoreCheckDirectoryName);
 
-        hr = StrAllocFormatted(&pwzNetCoreCheckFilePath, L"%ls%ls", pwzTempPath, L"NetCoreCheck.exe");
-        ExitOnFailure(hr, "failed to set NetCoreCheck file path");
+        hr = StrAllocFormatted(&pwzNetCoreCheckFilePath, L"%lsNetCoreCheck.exe", pwzNetCoreCheckDirectoryPath);
+        ExitOnFailure(hr, "failed to set NetCoreCheck file path for directory %ls", pwzNetCoreCheckDirectoryPath);
 
         hr = WcaExtractBinaryToFile(pwzNetCoreCheckBinaryId, pwzNetCoreCheckFilePath);
         ExitOnFailure(hr, "failed to extract NetCoreCheck from binary '%ls' to file %ls", pwzNetCoreCheckBinaryId, pwzNetCoreCheckFilePath);
 
-        // Read all NetCoreCheck.exe parameters and variable
+        // Read all NetCoreCheck.exe parameters and property
         hr = WcaGetRecordString(hRec, runtimeType, &pwzRuntimeType);
-        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.RuntimeType");
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.RuntimeType");
 
         hr = WcaGetRecordString(hRec, version, &pwzVersion);
-        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Version");
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.Version");
 
         hr = WcaGetRecordString(hRec, rollForward, &pwzRollForward);
-        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.RollForward");
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.RollForward");
 
-        hr = WcaGetRecordString(hRec, variable, &pwzVariable);
-        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Variable");
+        hr = WcaGetRecordString(hRec, property, &pwzProperty);
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.Property");
 
-        // Run NetCoreCheck.exe and store its result in variable
+        // Run NetCoreCheck.exe and store its result in property
         hr = StrAllocFormatted(&pwzCommandLine, L"-n %ls -v %ls -r %ls", pwzRuntimeType, pwzVersion, pwzRollForward);
         ExitOnFailure(hr, "failed to set NetCoreCheck command line");
         WcaLog(LOGMSG_VERBOSE, "Command: %ls %ls", pwzNetCoreCheckFilePath, pwzCommandLine);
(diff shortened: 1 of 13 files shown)


src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -449,6 +453,30 @@ static HRESULT FileIdExists(
...
+
+********************************************************************/
+static HRESULT GetNetCoreCheckBinaryId(
+    __in LPCWSTR pwzPlatform,
+    __out LPWSTR* ppwzBinaryId
+    )
+{
+    HRESULT hr = S_OK;
+
+    if (0 == lstrcmpW(L"x86", pwzPlatform)) {
```
robmen, line 467: nit: In WiX, curly braces go on their own line.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -449,6 +453,30 @@ static HRESULT FileIdExists(
...
+        hr = StrAllocString(ppwzBinaryId, L"Wix4NetCheck_x86", 0);
+    } else if (0 == lstrcmpW(L"x64", pwzPlatform)) {
```
robmen, line 469: `lstrcmpW` uses the user locale in the compare. In general, you'd want invariant compare `::CompareStringW(LOCALE_INVARIANT,`. And based on https://github.com/wixtoolset/issues/issues/6947 you might start the path towards `CompareStringOrdinal()` here.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+    LPWSTR pwzRollForward = NULL;
+    LPWSTR pwzVariable = NULL;
+    LPWSTR pwzCommandLine = NULL;
+
+    // Initialize
+    hr = WcaInitialize(hInstall, "DotNetCompatibilityCheck");
+    ExitOnFailure(hr, "failed to initialize");
+
+    // Get temp directory path
+    dwTempPathLength = GetTempPathW(dwTempPathLength, NULL);
```
robmen, line 941: Use `PathCreateTempFile()` from `pathutil.h`. It handles the string allocation for you. Pass `NULL` for `phTempFile` if you don't need the file handle.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+        hr = WcaGetRecordString(hRec, version, &pwzVersion);
+        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Version");
+
+        hr = WcaGetRecordString(hRec, rollForward, &pwzRollForward);
+        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.RollForward");
+
+        hr = WcaGetRecordString(hRec, variable, &pwzVariable);
+        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Variable");
+
+        // Run NetCoreCheck.exe and store its result in variable
```
robmen, line 994: @rseanhall just did something like this in Burn. Probably be good if he takes a look at it.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
```diff
@@ -5,23 +5,32 @@
   <?include ..\..\caDecor.wxi ?>
 
   <Fragment>
-    <CustomAction Id="$(var.Prefix)NetFxScheduleNativeImage$(var.Suffix)" DllEntry="SchedNetFx" Execute="immediate" Return="check" SuppressModularization="yes" BinaryRef="$(var.Prefix)NetFxCA$(var.Suffix)" />
+    <CustomAction Id="$(var.Prefix)NetFxScheduleNativeImage$(var.Suffix)" DllEntry="SchedNetFx" Execute="immediate" Return="ignore" SuppressModularization="yes" BinaryRef="$(var.Prefix)NetFxCA$(var.Suffix)" />
```
robmen, line 8: Why change this to `ignore`?

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
```diff
@@ -5,23 +5,32 @@
...
+
+    <InstallUISequence>
+      <Custom Action="$(var.Prefix)NetFxDotNetCompatibilityCheck$(var.Suffix)" Before="LaunchConditions" Overridable="yes" />
+    </InstallUISequence>
   </Fragment>
 
   <!-- NetFx Custom Action DLL Definitions -->
   <Fragment>
     <Binary Id="$(var.Prefix)NetFxCA$(var.Suffix)" SourceFile="!(bindpath.$(var.platform))netfxca.dll" />
+    <?foreach PLATFORM in x86;x64;arm64?>
```
robmen, line 32: Let's avoid using anything but variable substitution from the preprocessor unless really necessary. Better to duplicate the lines with the platform changing.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
```diff
@@ -5,23 +5,32 @@
...
+      <Binary Id="$(var.Prefix)NetCheck_$(var.PLATFORM)" SourceFile="!(bindpath.$(var.PLATFORM))NetCoreCheck.exe" />
```
robmen, line 33: Also, are these three platforms supposed to be included in each architecture of the .wixlib? Seems like this should just use the `$(var.Suffix)` pattern of the other custom actions here. But I could be missing something.

src/ext/NetFx/wixext/NetFxCompiler.cs
```diff
@@ -325,5 +361,184 @@ private void ParseNativeImageElement(Intermediate intermediate, IntermediateSect
                 });
             }
         }
+
+        /// <summary>
+        /// Parses a DotNetCompatibilityCheck element.
+        /// </summary>
+        /// <param name="element">The element to parse.</param>
+        private void ParseDotNetCompatibilityCheckElement(Intermediate intermediate, IntermediateSection section, XElement element)
```
robmen, line 369: Again, @rseanhall was deeper in the design of this feature. I think it'd be good if he evaluates the design.

robmen (changes requested):

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
StefanStojanovic: Previously, in NetFx extension, `NativeImage` was the only custom element. Its custom action fails if a view cannot be opened on `Wix4NetFxNativeImage` table. Now, there is a new custom element `DotNetCompatibilityCheck` whose custom action also fails if a view cannot be opened on its own table `Wix4NetFxDotNetCompatibilityCheck`. Using just one of the two elements would result with the other one's CA failing thus failing the entire installation, so using `ignore` for both made sense.

Another possibility is to make unsuccessful calls to `WcaOpenExecuteView` an early exit condition for the CA without failing it.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
StefanStojanovic: (Also relevant for [previous comment](https://github.com/wixtoolset/wix4/pull/262#discussion_r996262555)) The problem is that running a x86 version of NetCoreCheck.exe does not detect any of the x64 .NET runtimes and vice versa. I've also seen similar approach to this in Sean's change for Burn so I tried to replicate it as close as possible.

src/ext/NetFx/wixext/NetFxCompiler.cs
StefanStojanovic: I agree, let's see what Sean has to say about the PR and then I'll make all the changes requested by both of you.

src/ext/NetFx/ca/netfxca.cpp
robmen: Thinking about this more, I think we actually need to extract the exe to a randomly generated folder name. Then an attacker cannot attempt DLL hijack by simply dropping a file in `%TMP%`. `PathCreateTempDirectory()` with a random string should work.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
robmen: You'll want to use `WcaTableExists()` from `wcautil.h` to branch around these two code paths, instead.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
robmen: I see. Thanks for educating me.

This is unfortunate. .NET team has definitely pushed their detection problems up a layer, haven't they?

Still, do not use `foreach`.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+    while (S_OK == (hr = WcaFetchRecord(hView, &hRec)))
+    {
+        HANDLE hProcess = INVALID_HANDLE_VALUE;
+        DWORD dwExitCode = 0;
+
+        // Extract NetCoreCheck.exe for platform to temp directory
+        hr = WcaGetRecordString(hRec, platform, &pwzPlatform);
+        ExitOnFailure(hr, "failed to get Wix4DotNetCompatibilityCheck.Platform");
+
+        hr = GetNetCoreCheckBinaryId(pwzPlatform, &pwzNetCoreCheckBinaryId);
```
rseanhall, line 972: I'm not sure we need the extra complexity of this function when the id is simply 'Wix4NetCheck_' + pwzPlatform.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+        ExitOnFailure(hr, "failed to get NetCoreCheck binary id for platform %ls", pwzPlatform);
+
+        hr = StrAllocFormatted(&pwzNetCoreCheckFilePath, L"%ls%ls", pwzTempPath, L"NetCoreCheck.exe");
```
rseanhall, line 975: This could be precalculated outside the loop with this implementation.

I don't know if it's worth it, but there's a potential optimization here to extract each architecture to a different path so that the same file doesn't have to be extracted more than once. Which would mean this couldn't be precalculated anymore.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+        hr = ProcWaitForCompletion(hProcess, INFINITE, &dwExitCode);
+        ExitOnFailure(hr, "failed to finish NetCoreCheck from binary '%ls' with command line: %ls %ls", pwzNetCoreCheckBinaryId, pwzNetCoreCheckFilePath, pwzCommandLine);
+        WcaLog(LOGMSG_VERBOSE, "Exit code: %lu", dwExitCode);
+        CloseHandle(hProcess);
+
+        hr = WcaSetIntProperty(pwzVariable, dwExitCode);
+        ExitOnFailure(hr, "failed to set NetCoreCheck result in %ls", pwzVariable);
+
+        // Delete extracted NetCoreCheck.exe
+        DeleteFileW(pwzNetCoreCheckFilePath);
```
rseanhall, line 1011: Functions that are from a public SDK should be prefixed with the global namespace scope operator `::`.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+    }
+    if (E_NOMOREITEMS == hr)
+        hr = S_OK;
```
rseanhall, line 1014: Always use braces, even when optional.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +907,132 @@ extern "C" UINT __stdcall ExecNetFx(
...
+        ExitOnFailure(hr, "failed to set NetCoreCheck command line");
+        WcaLog(LOGMSG_VERBOSE, "Command: %ls %ls", pwzNetCoreCheckFilePath, pwzCommandLine);
+
+        hr = ProcExec(pwzNetCoreCheckFilePath, pwzCommandLine, SW_HIDE, &hProcess);
+        ExitOnFailure(hr, "failed to run NetCoreCheck from binary '%ls' with command line: %ls %ls", pwzNetCoreCheckBinaryId, pwzNetCoreCheckFilePath, pwzCommandLine);
+
+        hr = ProcWaitForCompletion(hProcess, INFINITE, &dwExitCode);
+        ExitOnFailure(hr, "failed to finish NetCoreCheck from binary '%ls' with command line: %ls %ls", pwzNetCoreCheckBinaryId, pwzNetCoreCheckFilePath, pwzCommandLine);
+        WcaLog(LOGMSG_VERBOSE, "Exit code: %lu", dwExitCode);
+        CloseHandle(hProcess);
```
rseanhall, line 1005: Normally we would use the `ReleaseHandle` macro. Also, the handle needs to be closed after `LExit` to make sure it isn't leaked if there is an error.

src/ext/NetFx/ca/netfxca.vcxproj
```diff
@@ -65,7 +65,16 @@
     <PackageReference Include="WixToolset.WcaUtil" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" PrivateAssets="All" />
     <PackageReference Include="GitInfo" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NET.Tools.NETCoreCheck.x86" GeneratePathProperty="true" />
+    <PackageReference Include="Microsoft.NET.Tools.NETCoreCheck.x64" GeneratePathProperty="true" />
+    <PackageReference Include="Microsoft.NET.Tools.NETCoreCheck.arm64" GeneratePathProperty="true" />
   </ItemGroup>
 
+  <Target Name="CopyNetCoreCheck" AfterTargets="Build">
```
rseanhall, line 73: I don't think these files should be copied anywhere. You should be able to add bind paths in the .wixproj for these locations and reference them that way. If you were using the hostfxr.dll as an example from my PR, then that was different because I needed it in the output directory to run the .exe from VS while developing it.

src/ext/NetFx/ca/precomp.h
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
 
 
+#include <string>
```
rseanhall, line 5: Was this necessary?

src/ext/NetFx/wixext/NetFxCompiler.cs
```diff
@@ -325,5 +361,184 @@ private void ParseNativeImageElement(Intermediate intermediate, IntermediateSect
...
+                                    runtimeType = "Microsoft.AspNetCore.App";
+                                    break;
+                                case "desktop":
+                                    runtimeType = "Microsoft.WindowsDesktop.App";
+                                    break;
+                                case "core":
+                                    runtimeType = "Microsoft.NETCore.App";
+                                    break;
+                                default:
+                                    this.Messaging.Write(ErrorMessages.IllegalAttributeValueWithLegalList(sourceLineNumbers, element.Name.LocalName, attrib.Name.LocalName, runtimeType, "aspnet, desktop and core"));
```
rseanhall, line 413: We should probably find a way to share code for these attributes that are shared with `DotNetCoreSearch`. If not, then the last parameter should be split up into `"aspnet", "desktop", "core"`.

src/ext/NetFx/wixext/NetFxCompiler.cs
```diff
@@ -325,5 +361,184 @@ private void ParseNativeImageElement(Intermediate intermediate, IntermediateSect
...
+                this.Messaging.Write(ErrorMessages.ExpectedAttribute(sourceLineNumbers, element.Name.LocalName, "Version"));
+            }
+
+            this.ParseHelper.ParseForExtensionElements(this.Context.Extensions, intermediate, section, element);
+
+            this.ParseHelper.CreateCustomActionReference(sourceLineNumbers, section, "Wix4NetFxDotNetCompatibilityCheck", this.Context.Platform, CustomActionPlatforms.ARM64 | CustomActionPlatforms.X64 | CustomActionPlatforms.X86);
+
+            if (!this.Messaging.EncounteredError)
+            {
+                // TODO: This will eventually be supported under Module/Product, but is not yet.
```
rseanhall, line 498: I assume that you actually meant supported under Bundle.

@robmen @barnson We haven't really discussed how this would work. My initial guess would be for it to work kind of like the `SoftwareTag` element. When the parent element is a `Fragment`, then `DotNetCompatibilityCheck` should either have the `Property` attribute (for Module/Product) or the `Variable` attribute (for Bundle). Both can't be specified. If it has the `Property` attribute, then the Bundle specific attributes (`Condition` and `After`) should not be specified.

If we go that direction, then for this PR it shouldn't be parsing `Variable`, `Condition`, or `After` and should be parsing `Property` instead.

src/ext/NetFx/wixext/Symbols/NetFxDotNetCompatibilityCheckSymbol.cs
```diff
@@ -0,0 +1,79 @@
...
+            NetfxSymbolDefinitionType.NetFxDotNetCompatibilityCheck.ToString(),
+            new[]
+            {
+                new IntermediateFieldDefinition(nameof(NetFxDotNetCompatibilityCheckSymbollFields.RuntimeType), IntermediateFieldType.String),
+                new IntermediateFieldDefinition(nameof(NetFxDotNetCompatibilityCheckSymbollFields.Platform), IntermediateFieldType.String),
+                new IntermediateFieldDefinition(nameof(NetFxDotNetCompatibilityCheckSymbollFields.Version), IntermediateFieldType.String),
+                new IntermediateFieldDefinition(nameof(NetFxDotNetCompatibilityCheckSymbollFields.RollForward), IntermediateFieldType.String),
+                new IntermediateFieldDefinition(nameof(NetFxDotNetCompatibilityCheckSymbollFields.Variable), IntermediateFieldType.String),
+            },
+            typeof(NetFxNativeImageSymbol));
```
rseanhall, line 20: typeof(NetFxDotNetCompatibilityCheckSymbol)

src/ext/NetFx/ca/netfxca.cpp
rseanhall: Ok, but MSI custom actions aren't my strong suit. This appears to be doing what it's supposed to: run Microsoft's NetCoreCheck.exe and store the exit code in the property that the user requested.

src/ext/NetFx/wixlib/NetFxExtension_Platform.wxi
StefanStojanovic: I stumbled upon a limitation when using `WcaTableExists()`. If table name is 32 characters or longer, it always returns `E_FAIL` because, as it seems, `::MsiDatabaseGetPrimaryKeysW()` cannot handle long names. The fix was easy - to shorten the table name, I just thought the limitation was worth noting.

src/ext/NetFx/ca/precomp.h
StefanStojanovic: At one point during the development yes, but not anymore. I'll revert it.

src/ext/NetFx/wixext/NetFxCompiler.cs
StefanStojanovic: As I saw, you're using enumerations for runtime type and platform. I could move to using that approach (plus add enumeration for roll forward policy) and then in CA instead of just reading values and passing them to NetCoreCheck.exe, I'd need to have map them back to values recognized by NetCoreChec.exe. In the end it comes down to which part of code do we want to have cleaner, so let me know your preference and I'll implement it that way.

src/ext/NetFx/wixext/NetFxCompiler.cs
StefanStojanovic: I actually copied the comment from `UtilCompiler.cs line 232` and my point was that if I uncomment the following line `this.ParseHelper.CreateWixSearchSymbol(...)`, while using `netfx:DotNetCompatibilityCheck` from within a `Product` I get the following: `warning WIX1150: The binder doesn't know how to place the following symbol into the output: SymbolName: 'WixSearch'`. So since my initial comment in code was ambiguous I hope it's now clear what I meant.

Second thing @rseanhall based on your comment on the issue https://github.com/wixtoolset/issues/issues/6264#issuecomment-1265710694 I've focused on MSI. I only added `Condition` and `After` because the design doc specified them, and as you can see, I'm already using `Variable` as a `Property`. So, since I've fixed most of the other comments you wrote on PR, once you decide how this should work let me know so I can make required changes and push everything for the second PR review iteration.

rseanhall: As far as I know, we don't currently have an element that can be authored in a fragment and that can either be used in an MSI or bundle which is why I included @robmen and @barnson. I was suggesting that the `Property` attribute would tell the compiler that it is for an MSI (don't create the search symbol) and that the `Variable` attribute would tell the compiler that it's for a bundle (do create the search symbol). The search symbol is really only for bundles. 

`Condition` and `After` are in the design doc because the initial plan was to have only one way to detect .NET Core. That meant that the element was going to be implemented for bundles, too. If the element doesn't actually use those attributes then it shouldn't be parsing them.

src/ext/NetFx/wixext/NetFxCompiler.cs
rseanhall: Maybe we'll try to get them to share code after this PR.

src/ext/NetFx/wixext/NetFxCompiler.cs
robmen: I agree with @rseanhall here. Given that this is focused on MSI, use the `Property` attribute only and skip the Bundle related attributes.

StefanStojanovic:
I've pushed changes made based on conversations we had here. All comments should be covered with these changes.

There are two comments ([this one](https://github.com/wixtoolset/wix4/pull/262#discussion_r996335296) and [this one](https://github.com/wixtoolset/wix4/pull/262#discussion_r996453520)) which I didn't follow completely. With latest changes, exporting is done via `PathCreateTempFile()`. I get a random file path (eg. `C:\Users\Mefi\AppData\Local\Temp\7D7E.tmp`) to which I export the binary, execute it and delete it afterwards. So, I haven't added any optimizations there as of now and I wanted to hear your thoughts on this. We can leave it as it is now, we can go with pre-exporting all architectures, etc. The optimal solution would probably be to go through all the records to get all listed platforms, parameters and properties and then export the needed binaries and run the commands, but this seems like an overengineered solution, especially if we'll mostly have one to few .NET compatibility checks per installer.

src/ext/NetFx/ca/netfxca.cpp
```diff
@@ -879,3 +892,124 @@ extern "C" UINT __stdcall ExecNetFx(
...
+    while (S_OK == (hr = WcaFetchRecord(hView, &hRec)))
+    {
+        // Extract NetCoreCheck.exe for platform to temp directory
+        hr = WcaGetRecordString(hRec, platform, &pwzPlatform);
+        ExitOnFailure(hr, "failed to get Wix4NetFxDotNetCheck.Platform");
+
+        hr = StrAllocFormatted(&pwzNetCoreCheckBinaryId, L"Wix4NetCheck_%ls", pwzPlatform);
+        ExitOnFailure(hr, "failed to get NetCoreCheck binary id for platform %ls", pwzPlatform);
+
+        hr = PathCreateTempFile(NULL, NULL, 0, L"", 0, &pwzNetCoreCheckFilePath, NULL);
```
robmen, line 948: Two things here:

1. We need to extract the binary to a randomly named directory to prevent the exe from being DLL hijacked. Since we don't control the exe this simple layer of protection seems prudent.

2.  Doesn't the file name need to be extracted with `.exe` extension so it can be executed later?

robmen (changes requested):

src/ext/NetFx/ca/netfxca.cpp
robmen: This is still an issue I called out in my last review.

StefanStojanovic: You're talking about [this comment](https://github.com/wixtoolset/wix4/pull/262#discussion_r1003656622) right? I was just waiting for confirmation on what to do there (whether to optimize with pre-exporting or not), I'll make the change to export binary to a random directory and push those changes.

src/ext/NetFx/ca/netfxca.cpp
StefanStojanovic: I've made the requested changes.

1. This directory has a randomly generated name from range NCC_0000 - NCC_7FFF.
2. Surprisingly, `.tmp` was working fine. Since there is now a randomly generated directory name, I went back to naming the extracted file `NetCoreCheck.exe`

robmen: 1. That isn't going to be random enough. The issue is that an attacker can seed all those directories in the System %TMP% folder with an attack DLL then wait for an elevated `NetCoreCheck.exe` to be run. I still recommend doing what Burn does and use a GUID (`GuidFixedCreate()` exists for that reason in `guidutil.h`). A GUID is _entirely_ non-guessable.

2. I am surprised `::CreateProcess()` cares less than I thought it would. 😄

StefanStojanovic: > 1. That isn't going to be random enough. The issue is that an attacker can seed all those directories in the System %TMP% folder with an attack DLL then wait for an elevated `NetCoreCheck.exe` to be run. I still recommend doing what Burn does and use a GUID (`GuidFixedCreate()` exists for that reason in `guidutil.h`). A GUID is _entirely_ non-guessable.

I agree, GUID was the first thing that came to my mind and functions provided by `guidutil.h` were the first thing I tried. However using either `GuidCreate` or `GuidFixedCreate` was resulting in LNK2019 error `error LNK2019: unresolved external symbol`

image

robmen: Ahh, I think you need to add `rpc.lib` (I'm away from my computer so that name isn't exactly it, but close, you will find it is used elsewhere) to the additional libraries in the vcxproj.

StefanStojanovic: Thanks a lot! Adding `rpcrt4.lib` worked. New changes pushed.

src/ext/NetFx/ca/netfxca.cpp
robmen: I 💯 agree, but since the most likely case is that this table has one row it isn't likely to be noticeable. This will be fine.

PR #265, opened 2022-10-16
Support add, modify and remove comments on user accounts

robmen:
Fixes wixtoolset/issues#5371
Closes #252


robmen:
@cpuwzd will you please review these changes? I squished your changes into a single commit. In doing so, I found a couple of minor bugs and fixed them in a separate commit here. Your PR wasn't open for maintainers to update it, so I created this one.

If this passes the build and you sign off, I think we'll be good to go.

cpuwzd:
I understand that it would take a few more cycles for me to fix these issues and that time is short, but I would appreciate a few notes on the motivation for your changes. Most of them seem to be calling conventions with regard to error returns. The better I understand the issues the better I can avoid them in the future.

robmen:
Here were the changes I applied:
1. If no error can be returned, `void` is fine
2. If an error can be returned, use `HRESULT` (exceptions to this are very rare)
3. Minimize/fix whitespace
4. Fixed handling of `\\` removal
5. Fixed the `SCAU_FAIL_IF_EXISTS` case so that it would fail with the correct error.

PR #267, opened 2022-10-19
Update dependencies and workaround ARM64 test failure

robmen:
* Update many NuGet packages
* Clean up shipping WiX .nupkgs
* Workaround ARM64 test failure until there is an ARM64 NETFX wix.exe - related to wixtoolset/issues#6960

### DIFF
--- a/src/internal/SetBuildNumber/Directory.Packages.props.pp
+++ b/src/internal/SetBuildNumber/Directory.Packages.props.pp
@@ -59,9 +59,9 @@
     <PackageVersion Include="Microsoft.VisualStudio.Setup.Configuration.Native" Version="1.14.114" />
     <PackageVersion Include="Microsoft.Win32.Registry" Version="4.7.0" />
 
-    <PackageVersion Include="NuGet.Credentials" Version="5.6.0" />
-    <PackageVersion Include="NuGet.Protocol" Version="5.6.0" />
-    <PackageVersion Include="NuGet.Versioning" Version="5.6.0" />
+    <PackageVersion Include="NuGet.Credentials" Version="6.3.1" />
+    <PackageVersion Include="NuGet.Protocol" Version="6.3.1" />
+    <PackageVersion Include="NuGet.Versioning" Version="6.3.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'" >
--- a/src/internal/SetBuildNumber/Directory.Packages.props.pp
+++ b/src/internal/SetBuildNumber/Directory.Packages.props.pp
@@ -45,29 +45,40 @@
     <PackageVersion Include="System.DirectoryServices" Version="4.7.0" />
     <PackageVersion Include="System.DirectoryServices.AccountManagement" Version="4.7.0" />
     <PackageVersion Include="System.IO.Compression" Version="4.3.0" />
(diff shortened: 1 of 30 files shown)


src/internal/SetBuildNumber/Directory.Packages.props.pp
```diff
@@ -45,18 +45,18 @@
...
     <PackageVersion Include="System.Reflection.Metadata" Version="1.6.0" />
     <PackageVersion Include="System.Security.Principal.Windows" Version="4.7.0" />
-    <PackageVersion Include="System.Text.Encoding.CodePages" Version="4.6.0" />
+    <PackageVersion Include="System.Text.Encoding.CodePages" Version="4.7.0" />
     <PackageVersion Include="System.Text.Json" Version="4.6.0" />
 
     <PackageVersion Include="Microsoft.AspNetCore.Owin" Version="3.1.13" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="3.8.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.3.1" />
```
rseanhall, line 56: Is this necessary? I vaguely remember needing to keep this lower to support VS2019, but that might not be correct.

src/internal/SetBuildNumber/Directory.Packages.props.pp
```diff
@@ -45,18 +45,18 @@
...
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit" Version="1.1.1" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.8.0" />
-    <PackageVersion Include="Microsoft.VisualStudio.Setup.Configuration.Native" Version="1.14.114" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.3.1" />
```
rseanhall, line 58: Same as above.

src/internal/SetBuildNumber/Directory.Packages.props.pp
rseanhall: This change will require users of WixToolset.Dnc.HostGenerator to have VS2022 17.3 or later - https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md.

robmen: Not strictly necessary. Is it necessary to keep it at `3.8.0` for VS2019 support?

rseanhall: According to that page, it needs to be 3.x to support VS2019.

rseanhall: It looks like VS2019 16.11 is the only supported version after this month so we should be able to go to `3.11.0`.

robmen: Weird. Your second comment didn't show up until I refreshed the page. That's a good doc link.

robmen: Updated to 3.11 and added the link to that documentation.

rseanhall: Looks good.

PR #268, opened 2022-10-19
Add queutil so Burn can manage its own queue of BA requested actions

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/6349


src/burn/engine/engine.cpp
```diff
@@ -864,52 +877,50 @@ static HRESULT RunApplication(
     // Unload BA.
     UserExperienceUnload(&pEngineState->userExperience, *pfReloadApp);
 
+    ::DeleteCriticalSection(&engineContext.csQueue);
+    ReleaseHandle(engineContext.hQueueSemaphore);
+    ReleaseQueue(engineContext.hQueue, FreeQueueItem, &engineContext);
```
robmen, line 882: Just a thought as I got this far: A critical section and semaphore seem useful for many uses of `que`. Did you consider allowing those to be initialized as part of creating a queue? Then deleting the queue handle could clean all those up together. They could also be optional in the initialization. Just a thought based on usage I've seen thus far.

src/libs/dutil/WixToolset.DUtil/queutil.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+    HRESULT hr = S_OK;
+    QUEUTIL_QUEUE_STRUCT* pQueue = NULL;
+
+    QueExitOnNull(phQueue, hr, E_INVALIDARG, "Handle not specified while creating queue.");
+
+    *phQueue = reinterpret_cast<QUEUTIL_QUEUE_HANDLE>(MemAlloc(QUEUTIL_QUEUE_HANDLE_BYTES, TRUE));
+    QueExitOnNull(*phQueue, hr, E_OUTOFMEMORY, "Failed to allocate queue object.");
+
+    pQueue = reinterpret_cast<QUEUTIL_QUEUE_STRUCT*>(*phQueue);
```
robmen, line 47: This doesn't seem necessary unless I'm missing something.

src/burn/engine/engine.cpp
rseanhall: I had the same idea but opted for simplicity. I think there are too many synchronization use cases to try to build it in and have a clean API.

src/libs/dutil/WixToolset.DUtil/queutil.cpp
rseanhall: You're right. I originally had the cleanup callback as a parameter to the create function here and stored it for when they eventually destroyed it. I need to remove this.

PR #281, opened 2022-11-10
WIXFEAT-3704 - Allow access to persisted variables from related bundles

jchoover:


jchoover:

PR #282, opened 2022-11-13
Start WiX Toolset v4-rc.1

robmen:


barnson:
Should global.json get an update?

PR #285, opened 2022-11-15
Go back to separate .wixlibs for DifxApp.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/7015.


src/ext/DifxApp/wixext/WixToolset.DifxApp.wixext.targets
```diff
@@ -7,5 +7,9 @@
        Condition=" Exists('$(MSBuildThisFileDirectory)..\$(WixToolsetExtensionPackageFolder)\$(MSBuildThisFileName).dll') " />
     <UnsupportedWixExtension Include="$(MSBuildThisFileName)"
       Condition=" !Exists('$(MSBuildThisFileDirectory)..\$(WixToolsetExtensionPackageFolder)\$(MSBuildThisFileName).dll') " />
+    <WixLibrary Include="$(MSBuildThisFileDirectory)..\$(WixToolsetExtensionPackageFolder)\difxapp_x86.wixlib"
```
robmen, line 10: What if we just put `$(Platform)` in the filename here? That would get us down to one `WixLibrary`. It would also fail with the appropriate error message when someone uses a `$(Platform)` that difx doesn't support.

src/ext/DifxApp/wixext/WixToolset.DifxApp.wixext.csproj
```diff
@@ -8,15 +8,13 @@
     <Description>WiX Toolset DIFxApp Extension</Description>
     <Title>WiX Toolset DIFxApp Extension</Title>
     <DebugType>embedded</DebugType>
+    <NuspecFile>$(MSBuildThisFileName).nuspec</NuspecFile>
```
robmen, line 11: You don't need to switch to .nuspec any more. Additional content can be included with `Content` items.

src/ext/DifxApp/wixext/WixToolset.DifxApp.wixext.csproj
```diff
@@ -8,15 +8,13 @@
...
   <Import Project="..\..\WixExt.props" />
 
-  <ItemGroup>
-    <EmbeddedResource Include="$(OutputPath)..\difxapp.wixlib" />
-  </ItemGroup>
-
   <ItemGroup Condition=" '$(NCrunch)'=='' ">
-    <ProjectReference Include="..\wixlib\difxapp.wixproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" />
+    <ProjectReference Include="..\wixlib\difxapp.wixproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Properties="Platform=x86" />
+    <ProjectReference Include="..\wixlib\difxapp.wixproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" Properties="Platform=x64" />
```
robmen, line 18: Add the attribute `OutputItemType="Content"` to the `ProjectReference` and the `difxapp_$(Platform).wixlib` should be added.

src/ext/DifxApp/wixext/WixToolset.DifxApp.wixext.targets
```diff
@@ -7,5 +7,9 @@
...
+       Condition=" Exists('$(MSBuildThisFileDirectory)..\$(WixToolsetExtensionPackageFolder)\difxapp_x86.wixlib') and '$(Platform)' != 'x64' " />
```
robmen, line 11: What if we remove the condition here so the build fails when uses a $(Platform) that difx doesn't support?

robmen (changes requested):

robmen (approved):

PR #288, opened 2022-11-17
Warn about unsuffixed custom action ids.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/7018.


src/wix/wix.cmd
```diff
@@ -24,9 +24,9 @@ msbuild -t:Restore wix.sln -p:Configuration=%_C% -nologo -m -warnaserror -bl:%_L
 
 
 :: Build
-msbuild wixnative\wixnative_t.proj -p:Configuration=%_C% -nologo -m -warnaserror -bl:%_L%\wixnative_build.binlog || exit /b
```
robmen, line 27: I thought this had been around for a long while and was not causing problems. It was only publish that seemed to have -m problems, no?

robmen (commented):
I don't think the changes to wix.cmd are necessary unless you saw failures in them.

src/wix/wix.cmd
barnson: Sorry, didn't mean to commit that. But yes, I'm having consistent build breaks with files-in-use errors.

robmen (approved):

PR #290, opened 2022-11-20
Latest fixes on path to -rc.1

robmen:
* Add support for UTF-8 console and use it for passing smartcab paths - fixes wixtoolset/issues#7024
* Allow signing targets to be overridden in simple Sdk-style .wixproj - closes wixtoolset/issues#7038
* Log error when path to executable cannot be found in MSBuild tool task - debugging wixtoolset/issues#7035

### DIFF
--- a/src/internal/WixInternal.BaseBuildTasks.Sources/BaseToolsetTask.cs
+++ b/src/internal/WixInternal.BaseBuildTasks.Sources/BaseToolsetTask.cs
@@ -3,6 +3,7 @@
 namespace WixToolset.BaseBuildTasks
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Runtime.InteropServices;
     using Microsoft.Build.Utilities;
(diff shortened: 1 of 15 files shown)


rseanhall (commented):
Wasn't the change to ConsoleInitialize to use UTF-8 a breaking change? Both sides of the pipe have to agree on what the format is, which is why you had to update the C# side, too. The .NET core detection exe needs to be updated to use UTF-8 as well, I think, due to this change.

robmen:
It would have been, but conutil was using the wrong codepage. That prevented non-ASCII from being read by conutil properly (the root of the bug). I went with UTF-8 because even when the codepage was fixed, Japanese/Chinese characters still could not be read because they didn't fit in the console codepage.

rseanhall:
I'm not sure you understood my point. The code page isn't wrong if the writer and the reader both use the same one. This changed the code page that the reader is using so now all users of conutil will need to make sure their writers are using the new one.

robmen:
I understood. The code page used by conutil was constant and incorrect. Switching to UTF-8 doesn't change the set of characters that were always correctly read _and_ is the codepage supported by the console that supports all characters.

rseanhall:
I understand that wixnative.exe needed to set it's console to UTF-8 to fix the bug. I don't understand why you think the old code page was incorrect in general. I don't think it was incorrect in general so I think this is a breaking change that probably needs to be done a little differently for backwards compatibility.

robmen:
The old code page used by conutil was incorrect in general and did not receive data outside of ASCII correctly. If conutil had used the correct codepage, then `ü` would have been transmitted correctly and the underlying bug would not have been found. Even with the correct console codepage, `道` did not transmit correctly. So, UTF-8 support could be added and everything that worked before continued to work (as much as it did) and UTF-8 could be used to access all characters.

conutil has a whole needs to be rewritten. In most cases, it only works with ASCII characters.

rseanhall:
This seems to be another instance where you have additional requirements in your head without filling me in. As far as I know, all that matters is the writer is using the same format as the reader. If you want to add more requirements then we can talk about that, but I didn't think conutil ever pretended to support anything past ASCII.

PR #297, opened 2022-12-07
Use MSI transaction end result to detect whether reboot is needed

nirbar:
I got logs showing that MsiEndTransaction may return ERROR_SUCCESS_REBOOT_REQUIRED
The proposed change allows default and custom BA to instruct the engine to reboot on these cases
Without this fix, the engine handles this return code as an error

Closes https://github.com/wixtoolset/issues/issues/7079


nirbar:

recheck

nirbar:

rseanhall (approved):
Thanks!

PR #304, opened 2022-12-13
Add message for DoAction WixUI custom actions.

barnson:


barnson:
Addresses remaining bit of https://github.com/wixtoolset/issues/issues/6998.

PR #331, opened 2023-01-16
Various fixes when using a .NET Core BA

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/7160
Fixes https://github.com/wixtoolset/issues/issues/7161
Fixes https://github.com/wixtoolset/issues/issues/7166

### DIFF
--- a/src/ext/Bal/wixext/BalErrors.cs
+++ b/src/ext/Bal/wixext/BalErrors.cs
@@ -43,6 +43,11 @@ namespace WixToolset.Bal
             return Message(sourceLineNumbers, Ids.IuibaPrimaryPackageEnableFeatureSelection, "When using WixInternalUIBootstrapperApplication, primary packages must not have feature selection enabled because it interferes with the user selecting feature through the MSI UI.");
         }
 
+        public static Message MissingDNCBAFactoryAssembly(SourceLineNumber sourceLineNumbers)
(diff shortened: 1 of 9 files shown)


src/ext/Bal/wixext/BalErrors.cs
```diff
@@ -43,6 +43,11 @@ public static Message IuibaPrimaryPackageEnableFeatureSelection(SourceLineNumber
             return Message(sourceLineNumbers, Ids.IuibaPrimaryPackageEnableFeatureSelection, "When using WixInternalUIBootstrapperApplication, primary packages must not have feature selection enabled because it interferes with the user selecting feature through the MSI UI.");
         }
 
+        public static Message MissingDNCBAFactoryAssembly(SourceLineNumber sourceLineNumbers)
+        {
+            return Message(sourceLineNumbers, Ids.MissingDNCBAFactoryAssembly, "The BA's entry point DLL must have bal:BAFactoryAssembly=\"yes\" when using the DotNetCoreBootstrapperApplicationHost.");
```
robmen, line 48: It would be better to use element names in the error messages to help direct users what to look for next. For example, instead of `BA` use `BootstrapperApplication` element and instead of just `entry point DLL` it would be good to mention the `Payload` element.

PR #332, opened 2023-01-16
Don't uninstall superseded MSI packages

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/7147


barnson:
This feels generally the right thing to do. Qs:

- Should the BA have an opportunity to override? In either direction?
- Should unsafe uninstall change this behavior?

rseanhall:
* The BA can request `BOOTSTRAPPER_REQUEST_STATE_FORCE_ABSENT` for the package. As long as there are no dependents, that will get it to uninstall.
* Probably not. Unsafe uninstall was not intended to force uninstall the bundle's packages. It was intended as a way to force the bundle to remove itself from ARP and the package cache.

barnson:
👍

PR #339, opened 2023-01-23
Fix Issue 7100

cpuwzd:
Fixes the specific problem raised in issue 7100.

Avoids duplicate calls to vs.FindVisualStudio.
Adds two new functions to ConversionLab.
Adds a few calls to ConversionLab.
Improves the appearance of the results of several conversion tests.
See the changes in the "expected" results of these tests.

### DIFF
--- a/src/wix/WixToolset.Converters/ConversionLab.cs
+++ b/src/wix/WixToolset.Converters/ConversionLab.cs
@@ -40,21 +40,58 @@ namespace WixToolset.Converters
             }
         }
 
+        public void InsertElementBeforeTargetElement(XElement newElement)
+        {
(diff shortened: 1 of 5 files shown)


cpuwzd:
The status of this pull request is:

Build Expected — Waiting for status to be reported

Is the CI build broken?

Ron

robmen:
PRs from non-maintainers require approval of a maintainer to run the build. This prevents the rampant bitcoin mining that was plaguing GitHub a while back. I've been busy catching up on other FireGiant work after spending so much time getting `-rc.2` out the door. It doesn't look like anyone else has had a chance to poke at this either.

I'll give the build a go but won't be back for a few more days.

PR #346, opened 2023-02-02
Read child processes stdout and stderr a-synch

nirbar:
See https://github.com/wixtoolset/issues/issues/7190

### DIFF
--- a/src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
+++ b/src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
@@ -21,6 +21,10 @@ static HRESULT CreatePipes(
 
     HRESULT hr = S_OK;
     SECURITY_ATTRIBUTES sa;
+    WCHAR wzGuid[GUID_STRING_LENGTH];
+    LPWSTR szStdInPipeName = NULL;
+    LPWSTR szStdOutPipeName = NULL;
+    BOOL fRes = TRUE;
     HANDLE hOutTemp = INVALID_HANDLE_VALUE;
     HANDLE hInTemp = INVALID_HANDLE_VALUE;
 
@@ -30,6 +34,16 @@ static HRESULT CreatePipes(
     HANDLE hInRead = INVALID_HANDLE_VALUE;
     HANDLE hInWrite = INVALID_HANDLE_VALUE;
 
+    // Generate unique pipe names
+    hr = GuidFixedCreate(wzGuid);
+    ExitOnFailure(hr, "Failed to create UUID.");
+
+    hr = StrAllocFormatted(&szStdInPipeName, L"\\\\.\\pipe\\%ls-stdin", wzGuid);
+    ExitOnFailure(hr, "Failed to create stdin pipe name.");
+
+    hr = StrAllocFormatted(&szStdOutPipeName, L"\\\\.\\pipe\\%ls-stdout", wzGuid);
+    ExitOnFailure(hr, "Failed to create stdout pipe name.");
+
     // Fill out security structure so we can inherit handles
     ::ZeroMemory(&sa, sizeof(SECURITY_ATTRIBUTES));
     sa.nLength = sizeof(SECURITY_ATTRIBUTES);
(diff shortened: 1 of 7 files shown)


src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
```diff
@@ -112,20 +132,73 @@ static HRESULT HandleOutput(
...
+                break;
+            }
+            ExitOnNullWithLastError((dwRes == ERROR_IO_PENDING), hr, "Failed to wait for stdout data");
+        }
+
+        dwRes = ::WaitForMultipleObjects(ARRAYSIZE(rghHandles), rghHandles, FALSE, INFINITE);
+        // Process terminated, or pipe abandoned
+        if ((dwRes == WAIT_OBJECT_0) || (dwRes == WAIT_ABANDONED_0) || (dwRes == (WAIT_ABANDONED_0 + 1)))
+        {
+            break;
```
rseanhall, line 190: If the user wants the output of the process, then this is not correct. It is possible that the process exits before all the output is read.

You'll probably need to make it wait only a specific amount of time when reading from the pipe. Then it would stop reading when either the process exited and it timed out or the pipe was broken. You should be able to use `AppWaitForSingleObject` with `ExitOnWaitObjectFailure` to help with knowing when it timed out.

src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
```diff
@@ -30,42 +38,51 @@ static HRESULT CreatePipes(
     HANDLE hInRead = INVALID_HANDLE_VALUE;
     HANDLE hInWrite = INVALID_HANDLE_VALUE;
 
+    // Generate unique pipe names
+    rs = ::UuidCreate(&guid);
```
rseanhall, line 42: Please use `GuidFixedCreate`

src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
```diff
@@ -30,42 +38,51 @@ static HRESULT CreatePipes(
...
     sa.lpSecurityDescriptor = NULL;
-
+    
     // Create pipes
-    if (!::CreatePipe(&hOutTemp, &hOutWrite, &sa, 0))
-    {
-        ExitOnLastError(hr, "Failed to create output pipe");
-    }
+    hOutTemp = ::CreateNamedPipeW(szStdOutPipeName, PIPE_ACCESS_INBOUND | FILE_FLAG_OVERLAPPED, PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT, 1, OUTPUT_BUFFER, OUTPUT_BUFFER, NMPWAIT_USE_DEFAULT_WAIT, &sa);
+    ExitOnNullWithLastError((hOutTemp && (hOutTemp != INVALID_HANDLE_VALUE)), hr, "Failed to create named pipe for stdout reader");
```
rseanhall, line 63: Please use the correct exit macros throughout. For example, this would be `ExitOnInvalidHandleWithLastError`.

src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
```diff
@@ -112,20 +132,73 @@ static HRESULT HandleOutput(
     BOOL bFirst = TRUE;
     BOOL bUnicode = TRUE;
     HRESULT hr = S_OK;
+    OVERLAPPED overlapped;
+    HANDLE rghHandles[2];
```
rseanhall, line 136: WiX style is to always initialize variables when they are declared.

src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
```diff
@@ -112,20 +132,73 @@ static HRESULT HandleOutput(
...
+    rghHandles[0] = hProcess;
+    rghHandles[1] = overlapped.hEvent;
+    
+    fRes = ::ConnectNamedPipe(hRead, &overlapped);
+    if (!fRes)
+    {
+        dwRes = ::GetLastError();
+        if (dwRes == ERROR_IO_PENDING)
+        {
+            dwRes = ::WaitForSingleObject(overlapped.hEvent, INFINITE);
```
rseanhall, line 158: Please use `AppWaitForSingleObject` with `ExitOnFailure`.

rseanhall (changes requested):

src/libs/wcautil/WixToolset.WcaUtil/qtexec.cpp
nirbar: Once the process has terminated, attempting to read from the pipe would yield ERROR_BROKEN_PIPE.
However, the process is bound to terminate after it has finished writing to the pipe, so I don't think that's really an issue

rseanhall: > attempting to read from the pipe would yield ERROR_BROKEN_PIPE

If that was true, then there would have been no issue. If the process does not close the handles then it will not yield ERROR_BROKEN_PIPE. Unless you're saying that the changes you made is now making the process close them?

> However, the process is bound to terminate after it has finished writing to the pipe

Yes, that's true. But that doesn't mean we have read all of the data from the pipe. If the user cares about the output, then we need to make sure we have read everything.

nirbar: See https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-exitprocess
Windows implicitly closes all the object handles of the process in step 6, before signaling the process object (step 8).
Thus by the time WaitForMultipleObjects returns with the process handle, the pipe had been closed already and would yield ERROR_BROKEN_PIPE if read from. This is also in line with tests I've been running

rseanhall: What's the point of reading asynchronously then? Trying to read synchronously from a broken pipe shouldn't hang.

nirbar: When the process returns with inheriting its pipes to other processes, then you're right.
However when the process inherits its pipes to other processes and exits, the wait on the pipe alone hangs. Waiting for either the process or the pipe fixes that

rseanhall: A synchronous read from a broken pipe will never hang. So if you're encountering an issue with the current code where it's hanging, then that means the pipe is not broken. I don't understand what you're trying to tell me. Either your changes have magically made it where the pipe now gets broken in which case we can keep the synchronous read. Or the pipe is still not being broken in which case we need to properly code the asynchronous reading such that the full output is captured. This code will not capture the full output because it gives up reading as soon as the process exits. There is no guarantee that we have read all the data from the pipe when the process exits because the OS can buffer the data in the pipe such that the code that is writing to the pipe can compete before the data is read.

nirbar: The problem arises when the process creates child processes of its own and they inherit the same stdout/stderr handles.
In that case even after the first process exits, the pipe handles are kept alive until the child processes exit as well. In practice instead the installation waits for the entire process tree.
The fix handles that case by waiting for either the process to terminate, or the pipe.

rseanhall: This function is doing two things. The first is waiting for the process to exit. Your updated code does this correctly. The second is capturing all of the output from that process. Your updated code does not do this.

nirbar: What makes you think that it doesn't capture the output? In my tests the output is logged the same as it was before the change. Did you see a test where the log was truncated?
Both the existing code and the PR check for ERROR_BROKEN_PIPE as a sign that the pipe has been closed on the write side, so it seems that both ways are at the same level of risk to miss out on some data. That said, I haven't seen logs where the process log was truncated

rseanhall: This is a classic race condition. It will work under some conditions and fail under other conditions. You should be able to force it to fail by sleeping at the beginning of the loop (line 166) and minimizing the size of the read buffer (`OUTPUT_BUFFER`).

nirbar: I agree with that. However, it is the same with the existing implementation
The PR doesn't increase the risk of loosing data, yet it fixes an indefinite hang in the described scenario

rseanhall: There is no race condition in the existing implementation. You have introduced it by breaking out of the loop when the process exited.

nirbar: I disagree with that- the same race condition exists implicitly
Once the process exits, any unread data in the pipe is discarded. 
You can see in the existing code that error ERROR_BROKEN_PIPE is silently ignored after ReadFile. You have no guarantee that there weren't some unlogged remains there.

rseanhall: No. The OS will continue to return the data that it buffered from a broken pipe. It will only return ERROR_BROKEN_PIPE when trying to read from an empty broken pipe.

nirbar: What makes you think that? Did you see it documented?

rseanhall: You can see the concept of buffering documented in https://learn.microsoft.com/en-us/windows/win32/ipc/anonymous-pipe-operations. It would be non-sensical for the buffer to be discarded so unless you can find documentation for that you should assume that it's actually possible to read all of the data from a pipe without having to synchronize the reader and writer to not close their end of the pipe before the other is done.

nirbar: I added a PeekNamedPipe call to read any remains in the pipe after the process has ended

rseanhall:
Sorry, I got distracted by the race condition in your original code and I missed the bigger picture. As we discussed in the [last WiX online meeting](https://www.firegiant.com/blog/2023/2/9/wix-online-meeting-251-highlights/):

1. We can't close the pipes because it could cause problems for the long running process if it tries to interact with them. Most processes will fail if they can't write to stdout or stderr, or can't read from stdin.
2. Even if we could do that, people might be relying on the current behavior where it is waiting for the pipes to be closed before it considers everything to be complete. The user will need to opt-in to this new functionality so we don't break existing code.

Based on that, I'm closing this PR since this approach is not acceptable. I think one possible alternate solution is to use the special file name `NUL` which is supposed to be like `/dev/null` in Unix. Calling `::CreateFile("NUL", ...)` should create a handle that allows the process to write as much as it wants without requiring a thread to read from it. It would have to be inherited in the child process like the anonymous pipe handles so that it doesn't get closed when the parent processes close.

P.S. `::PeekNamedPipe` introduces its own problems. It will overwrite the buffer from the async `::ReadFile`, and it doesn't remove the data that it read from the buffer, so there would be all kinds of data corruption.

`ExitOnNullWithLastError` should only be used with a single value that is nullable. For example:

```
fRes = ::ReadFile(hRead, pBuffer, OUTPUT_BUFFER - 1, nullptr, &overlapped);
if (!fRes)
{
    dwRes = ::GetLastError();
    if (dwRes == ERROR_BROKEN_PIPE)
    {
        break;
    }
    ExitOnNullWithLastError((dwRes == ERROR_IO_PENDING), hr, "Failed to wait for stdout data");
}
```

is normally written as:

```
if (!::ReadFile(hRead, pBuffer, OUTPUT_BUFFER - 1, nullptr, &overlapped))
{
    dwRes = ::GetLastError();
    if (ERROR_BROKEN_PIPE == dwRes)
    {
        break;
    }
    else if (ERROR_IO_PENDING != dwRes)
    {
        ExitWithLastError(hr, "Failed to wait for stdout data");
    }
}
```

PR #348, opened 2023-02-06
Add ability to use old style WIX Custom actions names

fk250046:
This is fix for issues #6821 and #7178.
Description:
If compiler faces old fasion name (prefixed with Wix4 to differ from WIX 3 actions) then suffix of platform is added. The action names are stored in the code. I have added action names that may be mentioned explicitly in a project, by my opinion.


robmen:
Thanks but no. The compiler should not have this level of internal knowledge of the utility extensions.

PR #350, opened 2023-02-08
Convert old style WIX Custom Actions names

fk250046:
Second try. (Issues #6821 and #7178)
Add ability to reference WIX custom actions without suffix.
Description:
To the BaseCompilerExtension class added new virtual property of type string list. Each extension compiler overrides this property with list of actual WIX custom action names without suffix. Core compiler reads this list while loading extension and uses it to add platform suffix to each Custom Action


robmen:
You are going about this the wrong way. Let's [discuss an actual desired solution](https://github.com/wixtoolset/issues/discussions/new?category=wix-development) before throwing code at random proposals.

PR #371, opened 2023-03-10
Warn when WiX is run on unsupported platforms

robmen:
Closes wixtoolset/issues#7267


barnson:
Should it not also warn via `dotnet build`?

robmen:
`dotnet build` will eventually call `wix.exe` through the `wix.targets`. I did think about adding a check somewhere early in `wix.targets,` but I didn't want to add more "fluff" to the build, so my hope is that all the tasks will run x-plat (they are pretty simple C#).

barnson:
+1

PR #375, opened 2023-03-20
Fix bug when parsing bal prereq attributes

rseanhall:
Fixes https://github.com/wixtoolset/issues/issues/7287

### DIFF
--- a/src/ext/Bal/test/WixToolsetTest.Bal/InternalUIBAFixture.cs
+++ b/src/ext/Bal/test/WixToolsetTest.Bal/InternalUIBAFixture.cs
@@ -51,6 +51,51 @@ namespace WixToolsetTest.Bal
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


rseanhall:
@robmen @barnson Are we still merging our own pull requests?

robmen:
Normally, yes, but given we're locking down for RTM, the new question is: Does the bug meet the RTM bar?

We haven't discussed the bar yet, so I don't know what the answer is at this moment. One option is to wait for Thursday to evaluate.

rseanhall:
That sure sounds like a no.

robmen:
I waffle (especially if there are unit tests to validate). @barnson tends to be more hardcore than me. 😄

barnson:
Unit tests add +1. There's a workaround, so -1. We'd need to add "known-and-dumb bugs" to track workarounds like this, so +0.5. So I'm net +0.5 to fix.

PR #376, opened 2023-03-23
Escape a trailing backslash in args to wixnative.exe

robmen:
Fixes wixtoolset/issues#7294

### DIFF
--- a/src/internal/WixInternal.TestSupport/DisposableFileSystem.cs
+++ b/src/internal/WixInternal.TestSupport/DisposableFileSystem.cs
@@ -35,7 +35,8 @@ namespace WixInternal.TestSupport
 
         public string GetFolder(bool create = false)
         {
-            var path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
+            // Always return a path with a space in it.
+            var path = Path.Combine(Path.GetTempPath(), ".WIXTEST " + Path.GetRandomFileName());
 
             if (create)
             {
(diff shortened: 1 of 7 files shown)


robmen:
Okay, @barnson and/or @rseanhall, this is ready for review (assuming the full battery of integration tests pass). I chose to add a space when generating test folders to flush out any other bugs with spaces. I only found problems in the test code itself.

barnson:
Good ratio of test:prod code! LGTM

rseanhall:
This is better than before but still not perfect. There's a reason I commented that ExternableExecutable code like I did - doing it properly is complex. The flaw here is that it's not verifying that the trailing backslash is escaped. If the input to `QuoteArgumentsAsNecesary` includes an arg `with spaces\\` then `"spaces\\"` is correct and `"spaces\\\"` is not. I haven't looked into whether we are standardizing the input to this function so maybe it doesn't matter.

rseanhall:
There is a correct native code implementation at https://github.com/wixtoolset/wix4/blob/0be5784115d2c8ea87c67bb4acedf8d545cf5615/src/libs/dutil/WixToolset.DUtil/apputil.cpp#L384

robmen:
We don't need the full command-line escaping solution for this code as it is internal, and we control its use. There could be more defense-in-depth to ensure folders with arbitrary counts of `\`s aren't added, but there is no way to slip them in through the codebase as it is (i.e. the paths always get normalized to remove multiple trailing slashes much earlier).

Thus this minimal change is sufficient.

rseanhall:
Then it sounds good to me

PR #378, opened 2023-03-25
Wix4 rc4 Patch creation outputs erroneous warning WIX1097

narcis-ro:
Fix for [Wix4 rc4 atch creation outputs erroneous warning WIX1097](https://github.com/wixtoolset/issues/issues/7315)


robmen:
Thanks for sending this. I'm taking this fix in a separate PR to include tests to ensure this does not regress in the future.

PR #379, opened 2023-03-25
Fix for wrong column index values in WixToolset.Dtf.WindowsInstaller.Linq.Entities.RemoveFile_

narcis-ro:
Fix for [](https://github.com/wixtoolset/issues/issues/7310)

Fixed also another issue in `Registry_`


robmen:
@narcis-ro thank you. Please sign the CLA as per the bot's instructions above. I'd also require changes to the DTF tests but those are not currently being executed since DTF Linq is an experimental feature that was never really completed. So, I'd take these changes (after you accept the CLA) because they clearly don't make the code worse.

narcis-ro:

PR #382, opened 2023-03-30
Set packagetype is MSBuildSdk in WixToolset.Sdk.nuspec

frg2089:
Nuget uses `MSBuildSdk` to indicate the SDK package type, so I think we should add that to the nuspec file.

They will show like this
image

- [Search packagetype=MSBuildSdk - Nuget.org](https://www.nuget.org/packages?packagetype=MSBuildSdk)
- [nuget packagetypes - Microsoft Docs](https://learn.microsoft.com/nuget/reference/nuspec#packagetypes)


robmen:
Good find. I just sent a [PR to get this package type documented](https://github.com/NuGet/docs.microsoft.com-nuget/pull/3036).

PR #384, opened 2023-03-30
Log registry root key upon search failures

icnocop:


src/burn/engine/search.h
```diff
@@ -95,6 +95,7 @@ typedef struct _BURN_SEARCH
             BURN_REGISTRY_SEARCH_TYPE Type;
             BURN_VARIANT_TYPE VariableType;
             HKEY hRoot;
+            LPWSTR sczRoot;
```
robmen, line 98: Rather than carry around duplicate data in a string variable only used for logging, what if we had a function that used a switch on the `hRoot` to return a constant string? There are several other functions that do the same for enums in Burn already so that would be a consistent pattern.

robmen (changes requested):

src/burn/engine/search.h
icnocop: I get the compiler error `expression must have integral or enum type` when trying to use a `switch` statement on an `HKEY`.

In any case, I updated the code to use a function to return the string representation of an `HKEY`.

Thank you.

PR #385, opened 2023-04-03
Handle error case when payload is missing both Name and SourceFile

robmen:
Fixes wixtoolset/issues#7333
Fixes wixtoolset/issues#7347


barnson (approved):
Oh so much better than it could've been...

rseanhall (approved):
Thanks! Technically speaking this is only for package elements, though. Everything else has `IsRequired` set to true so `CalculateAndVerifyFields` catches it there.

robmen:
Yes, exactly.

PR #386, opened 2023-04-04
Hide visible NetCoreCheck.exe window.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/7353.

Also adds unit test and removes dead code.


robmen (approved):
Okay, but I'm not deeply involved in NetCoreCheck. Would be good to get @rseanhall view on the removal of FeatureBand in particular.

barnson:
I'm not opposed to removing the cleanup (re-dirtying?).

rseanhall (approved):
Yeah, I don't know why feature band was added to NetCoreCheck.

PR #389, opened 2023-04-04
Fix build of outputs with comma in their filename

robmen:
Fixes wixtoolset/issues#7343


barnson (approved):
I'm amazed that Git also had no problems with the insanity of commas in filenames...

rseanhall (approved):
Does it make sense to quote the first argument, too?

PR #390, opened 2023-04-04
Fix missing table prefix.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/7360


rseanhall (approved):

robmen (approved):
This is the change that keeps giving.

PR #400, opened 2023-04-23
Fix build failure when bundle has multiple RemoteBundle elements

nirbar:
Fix https://github.com/wixtoolset/issues/issues/7429


robmen:
Thanks. I expect this issue to be triaged into v4.01, so could you enhance one of the unit tests to validate this field now? Alternatively, you could add the test case you attached to the bug, but adding an additional check to one of the Remote tests now seems like it would work.

Good catch on a subtle issue.

robmen:
PS: please add the test to the current commit. A single commit will make it easier to cherry-pick into v4.0.1.

PR #402, opened 2023-04-25
Fix WcaErrorMessage `cArgs==-1` case.

barnson:
Count args before creating message record.

Fixes https://github.com/wixtoolset/issues/issues/7422. 
Fixes https://github.com/wixtoolset/issues/issues/7444.


robmen (approved):
Fascinating. I do wonder how long has this been wrong.

barnson:
`cArgs==-1` is new to WiX v4.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -1024,6 +1024,8 @@ extern "C" UINT __stdcall CreateUser(
         ::CoUninitialize();
     }
 
+    WcaLog(LOGMSG_VERBOSE, "iAttributes = %d, er = %d, hr = 0x%08x", iAttributes, er, hr);
```
robmen, line 1027: This looks like debug information being left in. Is that message text useful in the context of the log file or could more be added to make it useful? Or should it be removed?

src/ext/Util/ca/scaexec.cpp
```diff
@@ -1033,6 +1035,8 @@ extern "C" UINT __stdcall CreateUser(
         er = ERROR_INSTALL_FAILURE;
     }
 
+    WcaLog(LOGMSG_VERBOSE, "iAttributes = %d, er = %d, hr = 0x%08x", iAttributes, er, hr);
```
robmen, line 1038: Same as above.

src/ext/Util/ca/scaexec.cpp
barnson: I'm adding additional logging to try to remote-debug the failure my change "introduced." I can't repro locally so I'm trying to narrow down possible causes. It will be removed when/if I fix the problem.

src/ext/Util/ca/scaexec.cpp
barnson: As above.

PR #416, opened 2023-06-02
ProjectReferences to wixlibs should participate in DefineConstants creation

robmen:
Also, fix documentation in ResolveWixLibraryReferences target.

Fixes wixtoolset/issues#7512


barnson (approved):
I usually mistrust one-liner fixes...

PR #417, opened 2023-06-07
Automatically add logging flag for any burn ExePackage, BundlePackage, and related bundles

nirbar:
Fixes wixtoolset/issues#5220: Automatically add logging flag for any burn ExePackage, BundlePackage, and related bundles


barnson:
Thanks!

PR #418, opened 2023-06-07
Extract detached containers with "wix burn extract"

nirbar:
Fix wixtoolset/issues#6298

### DIFF
--- a/src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
+++ b/src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
@@ -143,6 +143,65 @@ namespace WixToolset.Core.Burn.Bundles
         }
 
         /// <summary>
+        /// Extracts detached containers to the output directory.
+        /// </summary>
+        /// <param name="outputDirectory">Directory to write extracted files to.</param>
+        /// <param name="tempDirectory">Scratch directory.</param>
+        /// <param name="uxOutputDirectory">BA extraction folder. If null or empty, a BA folder will be created within tempDirectory</param>
+        /// <returns>True if successful, false otherwise</returns>
+        public bool ExtractDetachedContainers(string outputDirectory, string tempDirectory, string uxOutputDirectory)
+        {
+            if (String.IsNullOrEmpty(uxOutputDirectory))
+            {
+                uxOutputDirectory = Path.Combine(tempDirectory, "BA", "Final");
+            }
+
+            var manifestPath = Path.Combine(uxOutputDirectory, "manifest.xml");
+            if (!File.Exists(manifestPath))
+            {
+                var uxTempDirectory = Path.Combine(tempDirectory, "BA", "Temp");
+                if (!this.ExtractUXContainer(uxOutputDirectory, uxTempDirectory))
+                {
+                    return false;
+                }
+            }
+
+            var document = new XmlDocument();
+            document.Load(manifestPath);
+            var nsmgr = new XmlNamespaceManager(document.NameTable);
+            nsmgr.AddNamespace("burn", BurnCommon.BurnNamespace);
+            var detachedContainerNodes = document.SelectNodes("/burn:BurnManifest/burn:Container[not(@Attached = 'yes')]", nsmgr);
+
+            var exeDirectory = Path.GetDirectoryName(this.fileExe);
+            foreach (XmlElement containerElement in detachedContainerNodes)
+            {
+                var containerName = containerElement.GetAttribute("FilePath");
+
+                var cabinetPath = Path.Combine(exeDirectory, containerName);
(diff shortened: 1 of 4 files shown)


src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
```diff
@@ -142,6 +142,69 @@ public bool ExtractUXContainer(string outputDirectory, string tempDirectory)
...
+            var document = new XmlDocument();
+            document.Load(manifestPath);
+            var nsmgr = new XmlNamespaceManager(document.NameTable);
+            nsmgr.AddNamespace("burn", BurnCommon.BurnNamespace);
+            var detachedContainerNodes = document.SelectNodes("/burn:BurnManifest/burn:Container[not(@Attached = 'yes')]", nsmgr);
+
+            var exeDirectory = Path.GetDirectoryName(this.fileExe);
+            foreach (var node in detachedContainerNodes)
+            {
+                var containerElement = (XmlElement)node;
```
robmen, line 180: This cast can be avoided by changing the `var` in the `foreach` to `XElement`.

src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
```diff
@@ -142,6 +142,69 @@ public bool ExtractUXContainer(string outputDirectory, string tempDirectory)
...
+        /// </summary>
+        /// <param name="outputDirectory">Directory to write extracted files to.</param>
+        /// <param name="tempDirectory">Scratch directory.</param>
+        /// <param name="uxOutputDirectory">UX extraction folder. If null or empty, a UX folder will be created within tempDirectory</param>
+        /// <returns>True if successful, false otherwise</returns>
+        public bool ExtractDetachedContainers(string outputDirectory, string tempDirectory, string uxOutputDirectory)
+        {
+            if (string.IsNullOrEmpty(uxOutputDirectory))
+            {
+                uxOutputDirectory = Path.Combine(tempDirectory, "UX", "Final");
```
robmen, line 156: `UX` is a deprecated term. I know this is a temp folder, but let's not spread `UX` in more (possibly) user visible locations and use `BA` instead.

src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
```diff
@@ -142,6 +142,69 @@ public bool ExtractUXContainer(string outputDirectory, string tempDirectory)
...
+            {
+                var uxTempDirectory = Path.Combine(tempDirectory, "UX", "Temp");
+                if (!this.ExtractUXContainer(uxOutputDirectory, uxTempDirectory))
+                {
+                    return false;
+                }
+            }
+
+            Directory.CreateDirectory(tempDirectory);
+            Directory.CreateDirectory(outputDirectory);
```
robmen, line 170: I don't think we need to create either of these directories here. They will be created below only if necessary.

src/wix/WixToolset.Core.Burn/Bundles/BurnReader.cs
```diff
@@ -142,6 +142,69 @@ public bool ExtractUXContainer(string outputDirectory, string tempDirectory)
...
+
+                Directory.CreateDirectory(extractDirectory);
+                var cabinet = new Cabinet(cabinetPath);
+                cabinet.Extract(extractDirectory);
+
+                var containerId = containerElement.GetAttribute("Id");
+                var containerPayloadNodes = document.SelectNodes($"/burn:BurnManifest/burn:Payload[@Packaging='embedded' and @Container='{containerId}']", nsmgr);
+                foreach (var payloadNode in containerPayloadNodes)
+                {
+                    var payloadElement = (XmlElement)payloadNode;
```
robmen, line 194: Same as above, the cast can be avoided using the explicit type in the `foreach`.

src/wix/test/WixToolsetTest.CoreIntegration/BundleExtractionFixture.cs
```diff
@@ -23,7 +23,7 @@ public void CanExtractBundleWithDetachedContainer()
                 var exePath = Path.Combine(baseFolder, @"bin\test.exe");
                 var extractFolderPath = Path.Combine(baseFolder, "extract");
                 var baFolderPath = Path.Combine(extractFolderPath, "BA");
-                var attachedContainerFolderPath = Path.Combine(extractFolderPath, "WixAttachedContainer");
+                var msiPath = Path.Combine(extractFolderPath, "_1", "test.msi");
```
robmen, line 26: Interesting. What caused this to change? `WixAttachedContainer` is a better name than `_1` but I'm missing something.

robmen (changes requested):

PR #431, opened 2023-07-20
Add firewall extension decompiler, make firewall modifications work for msi changes

chrisbednarski:
Add firewall extension decompiler.

MSI database schema changes:
- made all wix table columns except for `Direction` nullable
- Changed wix table columns from numbers to strings to support using properties.

Compiler changes:
- the firewall element can be nested under ServiceInstall or util:ServiceConfig
- Added `OnUpdate` property to the Wix firewall language which affect what happens during MSI modifications and repairs:
  `EnableOnly` - enable existing firewall rules, do not change any other properties. This is similar to how wix V4 firewall CA works
  `DoNothing` - do not update any properties of existing firewall rules
  default: re-assign all properties on existing firewall rules

- Added columns for all firewall properties supported by Windows Vista/7/8+
[INetFwRule](https://learn.microsoft.com/en-us/windows/win32/api/netfw/nn-netfw-inetfwrule) Windows Vista / Server 2008
[INetFwRule2](https://learn.microsoft.com/en-us/windows/win32/api/netfw/nn-netfw-inetfwrule2) Windows 7 / Server 2008 R2
[INetFwRule3](https://learn.microsoft.com/en-us/windows/win32/api/netfw/nn-netfw-inetfwrule3) Windows 8/ Server 2012
The related COM interfaces are only queried when at least one of the properties of the firewall interface has been set by the wix compiler.
  
IE. If a firewall rule is created and none of the `INetFwRule2` properties are set, the firewall rule can be deployed on Vista/7/8+ . Same with `INetFwRule3` properties. If none are used, the firewall rule can be deployed on Vista/7/8+

When a `INetFwRule2` property is added to the firewall rule, it will no longer work on Vista, only on 7/8+
This is handled by the `feaAddINetFwRule2` and `feaAddINetFwRule3` flags in the `Attributes` column.


src/ext/Iis/ca/scacertexec.cpp
```diff
@@ -154,7 +154,7 @@ static HRESULT ExecuteCertificateOperation(
     LPWSTR pwzPFXPassword = NULL;
     LPWSTR pwzFilePath = NULL;
     BYTE* pbData = NULL;
-    DWORD cbData = 0;
+    DWORD_PTR cbData = 0;
```
chrisbednarski, line 157: VS was complaining about stack corruption

chrisbednarski:
Is it possible to unit test [this](https://github.com/wixtoolset/wix/pull/431/files#diff-c663bbad37136850739a550a0b83c703986a5481f22b634acb8a91684962f8c3R23) ?

I added some changes to enable nesting firewall exceptions under `ServiceInstall` and `util:ServiceConfig`

Looks like the built in testing framework only allows testing a [single extension](https://github.com/wixtoolset/wix/pull/431/files#diff-de7b983c8059020acd32915d87871ba4ab1e1ec1aaf5d884cf57ea2de67b98ebR40) at a time.

Would this have to be covered by an end to end test?

robmen:
> Would this have to be covered by an end to end test?

Unit tests can only validate that the correct data gets into the MSI. Integration (end-to-end) tests are required to execution of the CustomAction. There is such a test for some user scenarios.

barnson:
Or just add support for more than one extension. I'm sure this isn't the only place it could be used.

chrisbednarski:
> Or just add support for more than one extension. I'm sure this isn't the only place it could be used.

image

The firewall extension is referenced as a project in the `WixToolsetTest.Firewall`.
Making and using that change would imply adding `WiXExtension.Util` project to the `Firewall.wixext` solution

I don't know if that makes sense, or would be ok.

barnson:
It wouldn't be the only one; NetFx depends on both Bal and Util.

chrisbednarski:
> It wouldn't be the only one; NetFx depends on both Bal and Util.

??

I couldn't find any test projects referencing more than a single project they're testing.

image

barnson:
Ah, the production extension itself has the dependencies.

barnson:
It does use PackageReference, not ProjectReference, as I recall.

chrisbednarski:
> > Would this have to be covered by an end to end test?
> 
> Unit tests can only validate that the correct data gets into the MSI. Integration (end-to-end) tests are required to execution of the CustomAction. There is such a test for some user scenarios.

Where would be the best place to add integration tests for the firewall extension?

barnson:
src\test\msi\WixToolsetTest.MsiE2E. The Util.wixext tests live there so Firewall.wixext tests should be neighbors.

src/ext/Iis/ca/scacertexec.cpp
chrisbednarski: fixed by https://github.com/wixtoolset/wix/pull/438

chrisbednarski:
I will review this, add integration tests and rebase with the develop branch after Windows XP support has been removed from the firewall extension. Might be able to split this into multiple PRs so they're easier to review as well

barnson:
Because you're adding columns to the custom table, this will need a bit of triage discussion: I think we need to change the `Wix4` prefix to `Wix5`. Can you open a feature request issue about the features you want to add? That'll give me the opportunity to bring up the prefix (and mention for the historical record that XP support is dead-dead).

chrisbednarski:
No problem.

chrisbednarski:
Added more firewall attributes. inetfwrule1 and inetfwrule2 interface is all covered, but needs more E2E tests
inetfwrule3 still to come

chrisbednarski:
Would you be able to approve the build workflow on this?
I'd like to see if all the tests pass in your test environment.

Specifically, I'm not sure [how many network interfaces](https://github.com/minfos/wix/blob/feat/firewall-squashed/src/test/msi/WixToolsetTest.MsiE2E/FirewallExtensionTests.cs#L542) there will be on the build image.

chrisbednarski:
Woops! Let's try that again.

chrisbednarski:
Added more compiler/decompiler tests. I think I'm pretty much done code wise.
The PR description is way out of date for the moment though.

What's the next step to help get this merged ?

barnson:
At the moment, it's https://github.com/wixtoolset/wix/pull/451. I don't expect an immediate resolution, so if it's otherwise ready, you should look at building the v4 packages and checking them in.

chrisbednarski:
Is [this ](https://github.com/minfos/wix/tree/feat/firewall-squashed/src/test/msi/TestData/FirewallExtensionTests/CrossVersionMerge) what you mean?

barnson:
Looks right (probably don't need the .wixpdb). If we end up supporting v4 and v5 together, we could go back to building them individually, but it's likely we'll have Data or Extensibility changes that will prevent it.

chrisbednarski:
rebase + added E2E tests covering msi properties and OnChange flag

barnson:
Ready for review? Any other PRs need updating?

chrisbednarski:
Yes, ready for review. I've closed off some redundant PRs and rebased #451, which might still be useful

barnson:
Cool, thanks. I will gird my loins and jump in, probably next weekend.

barnson:
@chrisbednarski, can you apply this patch? I don't have permission to push to your branch. Checks out locally but I want the PR CI build, which I can't trigger for some reason. I'm hoping another commit will work.

[0001-Fix-typo-and-compiler-complaints.patch](https://github.com/wixtoolset/wix/files/13401860/0001-Fix-typo-and-compiler-complaints.patch)

chrisbednarski:
applied the patch

chrisbednarski:
and rebased with develop branch

barnson:
Very cool! Thanks for contributing (and working through everything).

barnson:
If you get very bored, the schema doc at https://github.com/wixtoolset/web/blob/master/src/xsd4/firewall.xsd could use doc on the new features.

PR #434, opened 2023-07-22
Add support for unit testing multiple extensions

chrisbednarski:
Added a constructor that takes an array of types, so integration between multiple extensions can be tested.

Fixed an error that showed up after adding the new constructor, because passing `null` as the 2nd argument was ambiguous.
ambiguous-constructor

Example package file:
image

Example unit test file:
image

### DIFF
--- a/src/internal/WixInternal.TestSupport/Builder.cs
+++ b/src/internal/WixInternal.TestSupport/Builder.cs
@@ -11,14 +11,29 @@ namespace WixInternal.TestSupport
         public Builder(string sourceFolder, Type extensionType = null, string[] bindPaths = null, string outputFile = null)
         {
             this.SourceFolder = sourceFolder;
-            this.ExtensionType = extensionType;
+            if (extensionType != null)
+            {
+                this.ExtensionTypes = new Type[] { extensionType };
+            }
+            else
+            {
+                this.ExtensionTypes = new Type[] { };
+            }
+            this.BindPaths = bindPaths;
+            this.OutputFile = outputFile ?? "test.msi";
+        }
+
+        public Builder(string sourceFolder, Type[] extensionTypes, string[] bindPaths = null, string outputFile = null)
+        {
+            this.SourceFolder = sourceFolder;
+            this.ExtensionTypes = extensionTypes;
             this.BindPaths = bindPaths;
             this.OutputFile = outputFile ?? "test.msi";
         }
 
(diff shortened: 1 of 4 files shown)


src/internal/WixInternal.TestSupport/Builder.cs
```diff
@@ -46,10 +57,13 @@ public string[] BuildAndQuery(Action<string[]> buildFunc, bool validate, params
                     "-intermediateFolder", intermediateFolder,
                 };
 
-                if (this.ExtensionType != null)
+                if (null != this.ExtensionType)
```
barnson, line 60: Yoda comparisons aren't needed in managed code. (They're still officially style for WiX C code, however.)

src/internal/WixInternal.TestSupport/Builder.cs
```diff
@@ -9,6 +9,17 @@ namespace WixInternal.TestSupport
...
+            this.SourceFolder = sourceFolder;
+            if (null != extensionType)
+            {
+                this.ExtensionType = new[] { extensionType };
+            }
+            this.BindPaths = bindPaths;
+            this.OutputFile = outputFile ?? "test.msi";
+        }
+
+        public Builder(string sourceFolder, Type[] extensionType , string[] bindPaths = null, string outputFile = null)
```
barnson, line 22: Extra space before comma.

src/internal/WixInternal.TestSupport/Builder.cs
```diff
@@ -18,7 +29,7 @@ public Builder(string sourceFolder, Type extensionType = null, string[] bindPath
 
         public string[] BindPaths { get; set; }
 
-        public Type ExtensionType { get; set; }
+        public Type[] ExtensionType { get; set; }
```
barnson, line 32: Make this plural.

barnson (changes requested):
Couple of minor tweaks.

src/internal/WixInternal.TestSupport/Builder.cs
chrisbednarski: thanks for the tip

src/internal/WixInternal.TestSupport/Builder.cs
chrisbednarski: updated

barnson (approved):

PR #438, opened 2023-08-02
Fix stack corruption around cbdata when adding a machine certificate

chrisbednarski:
Adding a machine certificate calls into `WcaReadStreamFromCaData(&pwz, &pbData, (DWORD_PTR*)&cbData);`

`WcaReadStreamFromCaData` writes to `DWORD_PTR` which is larger than the `DWORD cbData` allocated on the stack, so memory of `cbPFXPassword` is temporarily corrupted
Then `cbPFXPassword` is overwritten by a call to `WcaReadStringFromCaData(&pwz, &pwzPFXPassword)`

Capture1

Capture2

No stack corruption is reported after this change


barnson:
Thanks!

PR #439, opened 2023-08-07
Ensure the Font actions are scheduled for TrueType fonts

robmen:
TrueType fonts are denoted by the empty string in the FontTitle field of the FileSymbol. That means a non-null FontTitle field value means a font is being installed.

Fixes wixtoolset/issues#7593


barnson (commented):
🦑 or :shipit:

PR #443, opened 2023-08-13
Add integration tests for the firewall extenstion

chrisbednarski:
Examples of the output of a failing test (when the unique check is modified in the verifier self test)

image

image

image


chrisbednarski:
Is this heading in the right direction?

Is there anything I can do to help get this in or across the line?

What WiX version do you think firewall CA changes could be made in (after more integration tests have been added for the existing firewall CA functionality)?

chrisbednarski:
There is a substantial amount of code in the firewall CA for adding firewall rules on Windows XP and possibly prior to XP.
This is impossible to test on a newer OS without making changes to the CA.

Since Windows XP support ended a long time ago, should this be removed when making major changes in the future?

barnson:
At first glance, it looks like what I'd expect to see. I'll take a look, probably this weekend.

XP support is dead-dead so especially if there's extra code to keep it running.

The next feature release for WiX is v5 in the develop branch. v4.0.2 is for high-pri bug fixes only.

barnson:
Very cool, thanks!

PR #445, opened 2023-08-20
Use WcaGetRecordFormattedString to get firewall rule description

chrisbednarski:
- The [column definition](https://github.com/wixtoolset/wix/blob/develop/src/ext/Firewall/wixext/FirewallTableDefinitions.cs#L23) for the Description says it's a formatted string, but it was retrieved as a plain string
- fixed a few typos in the comments
- added missing RemoteAddresses and LocalAddresses assert
- added integration test for using formatted properties


chrisbednarski:
I've updated the failing test to use `Environment.GetEnvironmentVariable`

And AWS getting in my way, as per usual. Added `RemoteAddresses` to the firewall rule unique check.

barnson:
Thanks!

PR #446, opened 2023-08-21
Remove windows xp support from firewall extension

chrisbednarski:
- removed all old Windows XP code
- all integration tests pass


chrisbednarski:
rebased with develop

PR #447, opened 2023-08-25
Add firewall extension decompiler

chrisbednarski:
Split from #431


barnson:
Thanks for splitting!

PR #448, opened 2023-08-27
add FIREWALL_EXCEPTION_ATTRIBUTES struct

chrisbednarski:
refactoring, so function signatures don't have to be changed later on, when adding new firewall rule properties


src/ext/Firewall/ca/firewall.cpp
```diff
@@ -561,6 +549,16 @@ extern "C" UINT __stdcall ExecFirewallExceptions(
     pwz = pwzCustomActionData;
     while (pwz && *pwz)
     {
+        iTarget = fetUnknown;
+
+        ReleaseNullStr(attrs.pwzName);
```
barnson, line 554: This isn't necessary -- the StrUtil functions reuse/reallocate memory.

chrisbednarski: Updated the PR

chrisbednarski:
Hi, It would make my life easier when creating more PRs if this was merged in.

PR #449, opened 2023-08-27
expand protocol range of firewall extension, add tests

chrisbednarski:
Error message when compiling a firewall exception with a Port attribute specified, and the protocol value is not TCP or UDP

image

When using `netsh advfirewall firewall add rule`

image


chrisbednarski:
All revised and included in #431

PR #450, opened 2023-08-30
rename firewall extension table name to Wix5FirewallException

chrisbednarski:
merge after #449


chrisbednarski:
I see that I haven't taken [CUSTOM_ACTION_DECORATION](https://github.com/wixtoolset/wix/blob/develop/src/ext/caDecor.h) into account, so custom actions are still defined as V4 and not merging as required.

Any guidance on how to deal with this just for the Firewall extension?
Or should it be changed for all of V5 ?

chrisbednarski:
I have `Wix4` and `Wix5` firewall extension tables, but only a single set of custom actions (after suppressing WIX1055 and WIX1056)

image

barnson:
Because the firewall CAs suppress modularization GUIDs, they'll definitely need to be renamed. (Annoyingly, the CA DLL does _not_ have its modularization GUID suppressed. But it's been that way for about ever.) I lean toward changing the prefix only to match a table name change -- it's seems almost cruel to mix `Wix4` table names with `Wix5` custom action ids. I'll add this question to the issue for triage.

chrisbednarski:
I've updated the PR with so all three names are changed:
- table name
- CA name
- CA binary/DLL

image

The CAs of the v5 module (that matches the version of the MSI) is stomped over (expected)
This was done by suppressing warnings 1055 and 1056
image

Wix5 exception records are being merged as expected
image

Firewall records from the 4.0.1 module are safely in their own table
image

chrisbednarski:
Thanks for your reply. I'm sure this triage discussion will be interesting.

chrisbednarski:
I think I missed [this one](https://github.com/wixtoolset/wix/blob/develop/src/ext/Firewall/wixext/Symbols/FirewallSymbolDefinitions.cs#L15)

Will this have to be changed to 5.0.0?

barnson:
I don't think so: https://github.com/wixtoolset/wix/pull/455

But let's see what @robmen says.

chrisbednarski:
Test project files are included in #431, but tests run using binaries.
So this PR is redundant I think.

PR #451, opened 2023-09-01
Add assembly redirects for cross version integration tests

chrisbednarski:
I encountered a few issues before I was able to build V4 and V5 test projects together.

A module merge test will follow, but I thought this might be handy in a separate commit, in case it needs to be undone for some reason later on.

- `nuget.config` change loads `WixToolset.*` < v5 from nuget, but uses the local packages from the build folder for the v5 being built.
- assembly redirects are required when a project references other projects with a wildcard (eg. `<ProjectReference Include="**\*.wixproj" />` in [`TestData.proj`](https://github.com/wixtoolset/wix/blob/develop/src/test/msi/TestData/TestData.proj#L6)) and some of those projects are different versions.

**Before nuget.config change**
unabletofindpackage

**Before assembly redirects were added**
cannotload1
cannotload2

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -12,6 +12,7 @@
   <packageSourceMapping>
     <packageSource key="nuget">
       <package pattern="*" />
+      <package pattern="WixToolset.*" allowedVersions="[,{GitBaseVersionMajor})" />
     </packageSource>
     <packageSource key="build">
       <package pattern="WixToolset.*" />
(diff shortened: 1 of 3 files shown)


barnson:
FYI: This is a topic of some discussion. We need to determine (generally) how extensions are going to work in a mixed v4/v5 scenario. Please let me know if you get blocked by this and we can discuss alternatives (like temporarily committing .msis built with v4).

chrisbednarski:
Thanks. I should be ok .. I'll skip adding cross-version integration tests until a decision has been made.

chrisbednarski:
rebased with develop.
This might still provide value, pending your decision on wix versioning strategy.

My two cents: 
There seems to be no harm in having this assembly redirect regardless of how the wix versioning will work in the future.

PR #452, opened 2023-09-02
firewall extension cross version integration test

chrisbednarski:
includes #451

Hope this will be useful for the table name change triage

the PR creates an msi package with two modules
- `Module401` (4.0.1)
- `ModuleCurrent` (5.0.0 preview)

Because the msi itself contains two firewall rules, I had to suppress the warnings that were triggered by the CA of `ModuleCurrent`

The integration test installs the msi, verifies all the added firewall exceptions, then uninstalls them


chrisbednarski:
this one can be closed, because it's also included in #431

PR #458, opened 2023-09-17
Naked files and default feature

barnson:
Implements https://github.com/wixtoolset/issues/issues/7696.
Implements https://github.com/wixtoolset/issues/issues/7581.

`File` elements can appear where `Component` elements do in WiX v4. The
compiler generates an appropriate per-file component. Naked files under
`Directory`, `DirectoryRef`, `Fragment`, `StandardDirectory`, or
`Package` elements are included in a package via the [default-feature
feature](https://github.com/wixtoolset/issues/issues/7581). Naked files
appearing under `ComponentGroup`, `Feature`, `FeatureRef`, and
`FeatureGroup` generate the component and the reference to the parent
element.


barnson:
https://github.com/wixtoolset/wix/pull/479

PR #459, opened 2023-09-25
DotNetCompatibilityCheck: don't abort install on incompatible platform

apacker1:
If running NetCoreCheck.exe fails with error code ERROR_EXE_MACHINE_TYPE_MISMATCH or ERROR_BAD_EXE_FORMAT then don't abort the installation, just set the property to -1.  Fixes https://github.com/wixtoolset/issues/issues/7737

### DIFF
--- a/src/ext/NetFx/ca/netfxca.cpp
+++ b/src/ext/NetFx/ca/netfxca.cpp
@@ -977,12 +977,20 @@ extern "C" UINT __stdcall DotNetCompatibilityCheck(
         WcaLog(LOGMSG_VERBOSE, "Command: %ls %ls", pwzNetCoreCheckFilePath, pwzCommandLine);
 
         hr = ProcExec(pwzNetCoreCheckFilePath, pwzCommandLine, SW_HIDE, &hProcess);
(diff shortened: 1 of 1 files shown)


apacker1:

barnson:
Thanks. This looks good. We'll triage the issue at our meeting next week. My gut says `-1` is probably not the best value, just because of signs. I'm thinking 1 (+1) or 65535. Thoughts?

apacker1:
Yeah, I wasn't sure about that.  I guess the main thing is that the value shouldn't be something that the NetCoreCheck.exe authors might use in the future, which I don't really know how to predict.  However, https://github.com/dotnet/deployment-tools/blob/main/src/clickonce/native/projects/NetCoreCheck/NetCoreCheck.h does make it look like any additional error codes will probably just increment from the current ones that are defined, around 0x3000.  So 65535 should be safe.  Maybe they could conceivably use +1 if they come up with some "successful, but ..." result code?  So I guess I'd vote for 65535, but I'm happy to do whatever you think is best.

barnson:
Yeah, that was my thinking too. I left a comment on the issue so triage makes a recommendation.

barnson:
Triage team says `13` is a good compromise of return value.

apacker1:
@barnson OK, I've updated my code to use 13.  Also updated https://github.com/wixtoolset/web/pull/229 to reflect this change

barnson:
Thanks!

PR #460, opened 2023-10-08
Add tests to simulate issue 7739.

cpuwzd:
Generalize the issue to apply to most situations in which inner text is used in Wix 3 to represent condidtions. These all sufferfrom the mishandling of whitespace othere than the space character gets converted into an attribute value.

The XML parser generally treats all white space the same and reduces contiguous white space to a single space character.

All condition expressions are reduced to single line expressions. The user is free to reformat his conveerted code by inserting white space characters for the optimum readability readability.
The ne

### DIFF
--- a/src/wix/WixToolset.Converters/WixConverter.cs
+++ b/src/wix/WixToolset.Converters/WixConverter.cs
@@ -1040,7 +1040,7 @@ namespace WixToolset.Converters
                 {
                     var action = UppercaseFirstChar(xCondition.Attribute("Action")?.Value);
                     if (!String.IsNullOrEmpty(action) &&
-                        TryGetInnerText(xCondition, out var text, out comments, comments) &&
+                        TryGetInnerText(xCondition, out var text, out comments, true, false, comments) &&
                         this.OnInformation(ConverterTestType.InnerTextDeprecated, element, "Using {0} element text is deprecated. Use the '{1}Condition' attribute instead.", xCondition.Name.LocalName, action))
                     {
                         conditions.Add(new KeyValuePair<string, string>(action, text));
@@ -1237,9 +1237,14 @@ namespace WixToolset.Converters
             var message = element.Attribute("Message")?.Value;
 
             if (!String.IsNullOrEmpty(message) &&
-                TryGetInnerText(element, out var text, out var comments) &&
+                TryGetInnerText(element, out var text, out var comments, true, true) &&
                 this.OnInformation(ConverterTestType.InnerTextDeprecated, element, "Using {0} element text is deprecated. Use the 'Launch' element instead.", element.Name.LocalName))
             {
+                if (String.IsNullOrWhiteSpace(text))
+                {
+                    text = String.Empty;
+                }
+
                 using (var lab = new ConversionLab(element))
                 {
                     lab.ReplaceTargetElement(new XElement(LaunchElementName,
@@ -1983,8 +1988,14 @@ namespace WixToolset.Converters
 
             var xScript = xCustomAction.Attribute("Script");
 
-            if (xScript != null && TryGetInnerText(xCustomAction, out var scriptText, out var comments))
+            if (xScript != null && TryGetInnerText(xCustomAction, out var scriptText, out var comments, false, false, new List<XNode>()))
             {
+                if (null != scriptText)
+                {
+                    char[] whitespaceChars = { ' ', '\t', '\r', '\n' };
+                    scriptText = scriptText.Trim(whitespaceChars);
+                }
+
                 if (this.OnInformation(ConverterTestType.InnerTextDeprecated, xCustomAction, "Using {0} element text is deprecated. Extract the text to a file and use the 'ScriptSourceFile' attribute to reference it.", xCustomAction.Name.LocalName))
                 {
                     var scriptFolder = Path.GetDirectoryName(this.SourceFile) ?? String.Empty;
@@ -2150,10 +2161,7 @@ namespace WixToolset.Converters
                 var attribute = element.Attribute(attributeName);
                 if (attribute != null)
                 {
-                    if (!String.IsNullOrWhiteSpace(text))
-                    {
-                        this.OnError(ConverterTestType.InnerTextDeprecated, attribute, "Using {0} element text is deprecated. Remove the element's text and use only the '{1}' attribute. See the conversion FAQ for more information: https://wixtoolset.org/docs/fourthree/faqs/#converting-packages", element.Name.LocalName, attributeName);
-                    }
+                    this.OnError(ConverterTestType.InnerTextDeprecated, attribute, "Using {0} element text is deprecated. Remove the element's text and use only the '{1}' attribute. See the conversion FAQ for more information: https://wixtoolset.org/docs/fourthree/faqs/#converting-packages", element.Name.LocalName, attributeName);
                 }
                 else if (this.OnInformation(ConverterTestType.InnerTextDeprecated, element, "Using {0} element text is deprecated. Use the '{1}' attribute instead.", element.Name.LocalName, attributeName))
                 {
@@ -2854,20 +2862,45 @@ namespace WixToolset.Converters
             }
         }
 
-        private static bool TryGetInnerText(XElement element, out string value, out List<XNode> comments)
-        {
-            return TryGetInnerText(element, out value, out comments, new List<XNode>());
-        }
-
-        private static bool TryGetInnerText(XElement element, out string value, out List<XNode> comments, List<XNode> initialComments)
+        // This function is used to simultaneously extract the inner text of, and any comments embedded in, an XElement.
+        // The comments are returned in a list, so there can be several of them. In addition, the code is designed so that
+        // it can be called in a loop, processing one element at a time, extracting the comments cumulatively, and returning
+        // the inner text separately with each call. An initial list is passed into the "initialComments" parameter. This
+        // parameter defaults to null, which is replaced by an empty list. The augmented list is returned through the "comments"
+        // out parameter. Thiks list is then passed into the "initialComments" paameter the next time around the loop.
+        //
+        // The inner text is returned through the "value" out parameter. The function returns true if inner text is found.
+        //
+        // There are two other parameters that control the behavior of the function: The "protectQuotes" parameter, with a default
+        // value of "true", causes single and double quotes to be processed into an appropriate set of escape sequences (escaped with
+        // a backslash) ap sppropriate for a "Condition" attribute. Contiguous whitespace is coalesced into a single space character
+        // and leading and trailing whitespace is trimmed. When this paarameter is false, the inner text is returned with all whitespace
+        // intact.
+        //
+        // The other optional parameter, "reportWhitespace", has a default value of "false". When it is true, the value returned by the
+        // function is true, not only when non-whitespace text is returned in the "value" out parameter, but also if only whitespace was
+        // found. In this case, the text returned is a single space. If this parameter is false, the text found is trimmed of leading and
+        // trailing whitespace before being returned and the function only returns true if the returned text is not empty.
+        //
+        // There is one other important case in which the function returns true: If a CDATA node is processed, even if all the whitespace removed
+        // causes the remaining text to be empty, the function returns true. If in fact, the remaining text is empty, the text returned is a single
+        // space.
+        private static bool TryGetInnerText(XElement element,
+                                            out string value,
+                                            out List<XNode> comments,
+                                            bool protectQuotes = true,
+                                            bool reportWhitespace = false,
+                                            List<XNode> initialComments = null)
         {
             value = null;
-            comments = null;
-            var found = false;
-
+            char[] whitespaceChars = { ' ', '\t', '\r', '\n' };
             var nodes = element.Nodes().ToList();
-            comments = initialComments;
-            var nonCommentNodes = new List<XNode>();
+            comments = initialComments ?? new List<XNode>(); ;
+            char? currentQuote = null;
+            var inWhitespace = false;
+            var cDataFound = false;
+            var whitespaceFound = false;
+            var sb = new StringBuilder();
 
             foreach (var node in nodes)
             {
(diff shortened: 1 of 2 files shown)


src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,6 +2854,45 @@ private static void RenameElementToStandardDirectory(XElement element)
...
         }
 
+        // If TryGetInnerText succeeds, normalize the string value it returns by replacing all \r's \n's with spaces,
+        // then replacing all multiple spacws with single spaces. This is suitable for use in an XText representing a condition.
+        private static bool TryGetNormalizedInnerText(XElement element, out string value, out List<XNode> comments)
+        {
+            var found = TryGetInnerText(element, out var localValue, out comments);
+
+            if (found && null != localValue)
+
```
robmen, line 2864: nit: Unnecessary blank line.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,6 +2854,45 @@ private static void RenameElementToStandardDirectory(XElement element)
             }
         }
 
+        // If TryGetInnerText succeeds, normalize the string value it returns by replacing all \r's \n's with spaces,
+        // then replacing all multiple spacws with single spaces. This is suitable for use in an XText representing a condition.
+        private static bool TryGetNormalizedInnerText(XElement element, out string value, out List<XNode> comments)
+        {
+            var found = TryGetInnerText(element, out var localValue, out comments);
+
+            if (found && null != localValue)
```
robmen, line 2863: Usually use `String.IsNullOrEmpty` to check for empty string instead of `null`, unless there is something very special about null. So, `!String.IsNullOrEmpty(localValue)` would be better here.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,6 +2854,45 @@ private static void RenameElementToStandardDirectory(XElement element)
             }
         }
 
+        // If TryGetInnerText succeeds, normalize the string value it returns by replacing all \r's \n's with spaces,
+        // then replacing all multiple spacws with single spaces. This is suitable for use in an XText representing a condition.
```
robmen, line 2858: I think this is dangerous as it could replace double spaces in a quoted value with a single space, breaking the original intent. For example `PROP == "Foo  Bar"` where there are two spaces between Foo and Bar would get modified.

If anything, you could split the XML on `\r` and `\n`, removing empty lines then `.Trim()` the resulting values to cut the leading/trailing whitespace away then `.Join(' ')` the result back together.

But looping through each character required understanding the context for each use of inner text.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,6 +2854,45 @@ private static void RenameElementToStandardDirectory(XElement element)
...
+
+            if (found && null != localValue)
+
+            {
+                var whiteSpace = false;
+                var sb = new StringBuilder();
+
+                foreach (var c in localValue)
+                {
+                    if ('\u0020' == c || '\u000a' == c || '\u000d' == c)
```
robmen, line 2871: This is the complicated way of saying:

```C#
if (' ' == c || '\r' == c || '\n' == c)
```

Right? I think this is the wrong way to solve the problem (see comment above about Foo and Bar) but using Unicode values seems overkill, no?

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,6 +2854,45 @@ private static void RenameElementToStandardDirectory(XElement element)
...
+                    {
+                        if (!whiteSpace)
+                        {
+                            whiteSpace = true;
+                            sb.Append('\u0020');
```
robmen, line 2876: `sb.Append(' ');` ???

robmen (changes requested):

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: This is my style habit. No problem, I'll fix it.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: TryGetInnerText() returns null when there is no inner text, but it certainly won't do any harm to treat empty strings the same way I treat nulls. I'll fix it.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: I hadn't considered whitespace between quotes. Is non-blank whitespace explicitly disallowed by the grammar? Doe it, or should it produce an error? Let me give an example, using \r as carriage return, \n as new line, and \b as space.

<Condition>
\b\bvariable = "abc\b\b\r\nxyz"
<Condition/>

This would split the condition into

\b\bvariable = "abc\b\b


xyz"

Removing the empty lines then yields:

\b\bvariable = "abc\b\b
xyz"

Then, trimming both lines yields:

variable = "abc
xyz"

and joining with \n yields:

variable = "abc\nxyz"

This puts us back where we began without considering quotes, i.e. the XText constructor will create a value for Condition of "abc\u000axyz", which certainly won't do anything like what you might have had in mind.

I'm hoping that the non-space white space already generates an error when found in quotes, and that I will never see this construct. If it doesn't, then I'd like to produce an error (of what severity and through what messaging mechanism. Finally, I'd want to delete the offending characters and let the chips fall where they might.

I'm assuming that both species of quotes would be seen literally in the incoming inner text, Once stored as an attribute value, will the quotes need to be escaped or all converted to single quotes so that they can be used in an

attribute="value"

setting

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: Certainly, the Unicode encoding is overkill. I used it to emphasize the literal result of passing the literal characters through the XText constructor. This code is likely to disappear once I decide how to deal with the issue above.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: See previous issue.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: I thought that you were referring to line 2862 when I wrote my last comment. Line 2864 is simply an error. That's the line I will remove.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: I have rewritten TryGetNormalizedInnerText(). Highlights are:

- Replace contiguous unquoted whitespace with a single space.
- Use single quotes for all outgoing outer quotes.
- Escape outgoing nested quotes with backslash.
- Escape all incoming backslashes with backslash.
- Trim the resulting string.

I hope I haven't overlooked anything.

cpuwzd:
Assuming that this pull request is still active, I'm ready for another review. This issue turned out to be considerably larger than first thought. The issue boils down to, "What do you want to know about the inner text?" The processing of inner text has been rethought in every scenario I could come up with. We could still use some more realistic test cases.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,39 +2862,168 @@ private static void RenameElementToStandardDirectory(XElement element)
...
+                                }
+                            }
+                            else
+                            {
+                                inWhitespace = false;
+
+                                if ('\\' == c)
+                                {
+                                    // Escape backslash.
+                                    nodeSB.Append('\\');
```
robmen, line 2951: Why is a backslash escaped? Backslashes aren't escaped in XML. Is there some other reason?

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,39 +2862,168 @@ private static void RenameElementToStandardDirectory(XElement element)
...
+                                }
+
+                                else if (!isCData)
+                                {
+                                    // Escape nested quote.
+                                    if ('\'' == c || '"' == c)
```
robmen, line 2957: Single and double quote are not escaped using backslashes in XML. Is this getting used for something else? Since we're using an XAttribute to set the value, we don't have to escape anything. In fact, escaping would cause values to be double escaped.

Am I missing why these need to be backslash escaped?

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,39 +2862,168 @@ private static void RenameElementToStandardDirectory(XElement element)
...
         {
             value = null;
-            comments = null;
-            var found = false;
-
+            char[] whitespaceChars = { ' ', '\t', '\r', '\n' };
             var nodes = element.Nodes().ToList();
-            comments = initialComments;
-            var nonCommentNodes = new List<XNode>();
+            comments = initialComments ?? new List<XNode>(); ;
```
robmen, line 2898: Nit: Remove trailing redundant `;`

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2854,39 +2862,168 @@ private static void RenameElementToStandardDirectory(XElement element)
...
+        // function is true, not only when non-whitespace text is returned in the "value" out parameter, but also if only whitespace was
+        // found. In this case, the text returned is a single space. If this parameter is false, the text found is trimmed of leading and
+        // trailing whitespace before being returned and the function only returns true if the returned text is not empty.
+        //
+        // There is one other important case in which the function returns true: If a CDATA node is processed, even if all the whitespace removed
+        // causes the remaining text to be empty, the function returns true. If in fact, the remaining text is empty, the text returned is a single
+        // space.
+        private static bool TryGetInnerText(XElement element,
+                                            out string value,
+                                            out List<XNode> comments,
```
robmen, line 2890: `out` values should go at the end of the function. Also, we don't wrap parameter lines like this in WiX.

The multiple `out` values suggest that this function is either doing too much and needs to be split, or it should return an object with the data instead of three return parameters.

src/wix/WixToolset.Core/IOptimizer.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
```
robmen, line 1: You need to rebase your change so that this change is not mixed in the middle.

robmen (changes requested):

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: My misunderstanding. Code removed.

src/wix/WixToolset.Converters/WixConverter.cs
cpuwzd: All out parameters removed and replaced by use of a context  parameter.

src/wix/WixToolset.Core/IOptimizer.cs
cpuwzd: Done.

robmen (changes requested):
Sorry, I got really sick for the last couple weeks so I didn't get to the review until today.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -1080,15 +1081,17 @@ private void ConvertComponentElement(XElement element)
             var xCondition = element.Element(ConditionElementName);
             if (xCondition != null)
             {
-                if (TryGetInnerText(xCondition, out var text, out var comments) &&
+                var context = new InnerTextContext();
+
+                if (TryGetInnerText(xCondition, context) &&
```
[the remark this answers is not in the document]
cpuwzd: Upon further review, it is not clear whether you are suggesting that the new context object being returned be a copy of the context object being passed in, with changes made as necessary. This would leave the option as to whether the caller overwrites the the variable referring to the input context. Using different variables might be useful in some scenario where the original context would be reused.

You suggest later that the comment collection be initialized and exposed by the context's constructor. That seems to imply that comments recovered by previous iterations of a loop be added to a new collection each time around that loop. The current code avoids copying those comments.

This also raises the question of, "Who owns the collection?"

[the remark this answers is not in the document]
cpuwzd: I decided to take a different approach. Read on.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2975,6 +3079,44 @@ private static bool WasImplicitlyStringTyped(string value)
...
+                {
+                    if (null == this.comments)
+                    {
+                        this.comments = new List<XNode>();
+                    }
+
+                    return this.comments;
+                }
+
+                set => this.comments = value;
```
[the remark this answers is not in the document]
cpuwzd: I think that I have this covered.

src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -2975,6 +3079,44 @@ private static bool WasImplicitlyStringTyped(string value)
...
+        //
+        // The Comments property always returns a (possibly empty) List<XNode> representing any comments that have been accumulated from previous calls
+        // to TryGetInnerText(), using the same InnerTextContext object. Even if the intended call to TryGetInnerText is short-circuited, the Comments
+        // property still returns the intended List<XNode>.
+        //
+        // Upon entry to TryGetInnerText(), the CoalesceWhitespace and reportWhitespace properties contain values that control the behavior of TryGetInnerContext(),
+        // as explained in the introductory comments of that function.
+        private class InnerTextContext
+        {
+            internal string Value { get; set; } = null;
```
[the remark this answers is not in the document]
cpuwzd: I have this covered.

PR #466, opened 2023-11-07
Include duplicated inline directory symbols referenced in subsequent sections

robmen:
Due to the handling of redundant symbols, which are only used by inline directory syntax, the symbols were only defined in the first section encountered by the linker. Fix that so at most one duplicated inline directory symbol is included when referenced.

Fixes wixtoolset/issues#7840

### DIFF
--- a/src/wix/WixToolset.Core/Link/FindEntrySectionAndLoadSymbolsCommand.cs
+++ b/src/wix/WixToolset.Core/Link/FindEntrySectionAndLoadSymbolsCommand.cs
@@ -42,13 +42,13 @@ namespace WixToolset.Core.Link
         /// Gets the collection of redundant symbols that should not be included
         /// in the final output.
         /// </summary>
-        public ISet<IntermediateSymbol> RedundantSymbols { get; private set; }
+        public ISet<IntermediateSymbol> IdenticalDirectorySymbols { get; private set; }
 
         public void Execute()
         {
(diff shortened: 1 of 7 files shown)


src/wix/WixToolset.Core/Link/FindEntrySectionAndLoadSymbolsCommand.cs
```diff
@@ -76,28 +76,27 @@ public void Execute()
...
                     {
-                        symbolsByName.Add(symbolWithSection.Name, symbolWithSection);
+                        symbolsByName.Add(fullName, symbolWithSection);
                     }
                     else // uh-oh, duplicate symbols.
                     {
                         // If the duplicate symbols are both private directories, there is a chance that they
-                        // point to identical symbols. Identical directory symbols are redundant and will not cause
-                        // conflicts.
+                        // point to identical symbols. Identical directory symbols should be treated as redundant\
```
barnson, line 92: trailing `\`

barnson (approved):
Comment typo, other lgtm

src/wix/WixToolset.Core/Link/FindEntrySectionAndLoadSymbolsCommand.cs
robmen: Yep. Fixed typo and force pushed.

PR #471, opened 2023-11-20
ArpEntry reads QuietUninstallString or UninstallString, and uses UninstallArguments for the uninstall command line

nirbar:
Fixes wixtoolset/issues#7877


barnson:
Thanks!

PR #473, opened 2023-12-01
Resolve issues with inner text

cpuwzd:
Fixes wixtoolset/issues#7739


robmen:
Yeah, I really like this refactor.

PR #478, opened 2023-12-24
Add default major upgrade.

barnson:
Add Package/@UpgradeStrategy to allow `none` to suppress major upgrade.

Implements https://github.com/wixtoolset/issues/issues/7605.

Requires https://github.com/wixtoolset/wix/pull/435.

### DIFF
--- a/src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
+++ b/src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
@@ -2,6 +2,7 @@
 
 namespace WixToolset.Core.Link
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using WixToolset.Data;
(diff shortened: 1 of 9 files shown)


src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+                {
+                    Table = "Directory",
+                    PrimaryKeys = WixStandardInstallFolderParent,
+                });
+            }
+
+            var upgradeSymbols = symbols.OfType<UpgradeSymbol>();
+            if (!upgradeSymbols.Any())
+            {
+                var packageSymbol = symbols.OfType<WixPackageSymbol>().SingleOrDefault();
```
robmen, line 63: `SingleOrDefault()` will force the enumeration to go through all symbols, which could be slightly expensive for really big builds. I would use `FirstOrDefault()` and let the binder deal with making sure there is a single package symbol.

Also, the `WixPackageSymbol` should be in the `EntrySection` so it could limit the search to just that section rather than all symbols. But I think the EntrySection is the first in the list of sections so a `FirstOrDefault()` across all sections will probably perform the same.

src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
```diff
@@ -0,0 +1,136 @@
...
+
+                if (packageSymbol?.UpgradeStrategy == WixPackageUpgradeStrategy.MajorUpgrade
+                    && !String.IsNullOrEmpty(packageSymbol?.UpgradeCode))
```
robmen, line 66: I wonder if we should always add the default major upgrade symbols if the strategy is `majorUpgrade`? I suppose the issue is that `majorUpgrade` strategy is default in v5 and people probably have `MajorUpgrade` elements from v4 which would create a collision. :hurtrealbad:

robmen (changes requested):

src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
barnson: `if the strategy is majorUpgrade` and there are no UpgradeSymbols so an existing MajorUpgrade won't trigger this.

robmen (approved):

PR #479, opened 2023-12-29
Support naked files.

barnson:
Implements https://github.com/wixtoolset/issues/issues/7696.

`File` elements can appear where `Component` elements do in WiX v4. The compiler generates an appropriate per-file component. Naked files under `Directory`, `DirectoryRef`, `Fragment`, `StandardDirectory`, or `Package` elements are included in a package via the [default-feature feature](https://github.com/wixtoolset/issues/issues/7581). Naked files appearing under `ComponentGroup`, `Feature`, `FeatureRef`, and `FeatureGroup` generate the component and the reference to the parent element.

Components and naked Files default to being installed to INSTALLFOLDER
(including a default INSTALLFOLDER if one isn't otherwise authored).


barnson:
Added:

Components and naked Files default to being installed to INSTALLFOLDER
(including a default INSTALLFOLDER if one isn't otherwise authored).

src/wix/WixToolset.Core/Compiler.cs
```diff
@@ -5485,6 +5456,227 @@ private YesNoType ParseFileElement(XElement node, string componentId, string dir
...
+        /// <param name="sourcePath">Default source path of parent directory.</param>
+        /// <returns>Yes if this element was marked as the parent component's key path, No if explicitly marked as not being a key path, or NotSet otherwise.</returns>
+        private void ParseNakedFileElement(XElement node, ComplexReferenceParentType parentType, string parentId, string directoryId, string sourcePath)
+        {
+            var sourceLineNumbers = Preprocessor.GetSourceLineNumbers(node);
+            var win64 = this.Context.IsCurrentPlatform64Bit;
+            string condition = null;
+            string subdirectory = null;
+
+            var keyPath = this.ParseFileElementAttributes(node, "TemporaryComponentId", directoryId, diskId: CompilerConstants.IntegerNotSet, sourcePath, out var _, componentGuid: "*", isNakedFile: true, out var fileSymbol, out var assemblySymbol);
```
robmen, line 5593: How about setting this to an invalid Id like `"@WixTempComponentId"`? That way, if it ever sneaks through too far into the build process, an error will occur.

robmen (commented):
Overall, looks good. One small suggestion.

src/wix/WixToolset.Core/Compiler.cs
barnson: Ooh, I like that!

PR #485, opened 2024-01-07
Extensions can create and extract containers using bundle extensions.

nirbar:
Fixes wixtoolset/issues#4889


barnson:
Please open a new issue for this feature. 4889 is specifically about pulling container logic into DUtil. I'm not sure I see the value in making this extensible, compared to just adding LZMA support into the engine.

robmen:
Please do start with a feature request so we can discuss the feature and its design to decide if it is a good candidate for the WiX toolset. That will save you a lot of time and re-implementation of features if the design is changed significantly.

PR #488, opened 2024-01-20
Fix crash when CloseApplications_X64 closes Electron app

avjts:
CloseApplications custom action may crash on 64-bit systems if the number of processes to close exceeds the default memory alignment for a platform. E.g. on AMD64 seven processes is enough to trigger the problem, while on ARM64 it has to be more.

I encountered the above problem trying to close an Electron application. Electron apps have at least 6 child processes of the very same name as the master process. It causes a buffer overflow in `ProcWaitForIds` function.


avjts:

avjts:
recheck

barnson:
Thanks!

PR #495, opened 2024-02-18
WiX v5 uses `wixext5` NuGet package subdirectory.

barnson:


barnson:
Addresses https://github.com/wixtoolset/issues/issues/7997.
Addresses https://github.com/wixtoolset/issues/issues/7709.

src/wix/WixToolset.Core.ExtensionCache/ExtensionCacheManager.cs
```diff
@@ -29,12 +30,17 @@ public ExtensionCacheManager(IMessaging messaging, IExtensionManager extensionMa
         {
             this.Messaging = messaging;
             this.ExtensionManager = extensionManager;
+
+            var fileVersion = FileVersionInfo.GetVersionInfo(typeof(ExtensionCacheManager).Assembly.Location);
+            this.WixVersion = fileVersion.ProductMajorPart.ToString();
```
robmen, line 35: Another option is to use `typeof(ExtensionCacheManager).Assembly.GetType().Name.Version.Major` (or something a lot like that). That gives the AssemblyVersion instead of FileVersion.

robmen (approved):
One option to consider.

PR #497, opened 2024-02-27
Use camelCase attribute values.

barnson:


src/ext/Firewall/wixext/FirewallDecompiler.cs
```diff
@@ -499,4 +501,14 @@ private void FinalizeFirewallExceptionTable(TableIndexedCollection tables)
             }
         }
     }
+
+    internal static class StringExtensions
+    {
+        public static string ToCamelCase(this string str)
+        {
+            var camelCase = str[0].ToString().ToLowerInvariant() + str.Substring(1);
```
robmen, line 509: Should this protect against `null` and `String.Empty`?

robmen (commented):
One small thing to consider in camelCase.

src/ext/Firewall/wixext/FirewallDecompiler.cs
barnson: I'd be more worried about something public, because surely I will never make such a mistake...

PR #505, opened 2024-03-12
Fix localization collation.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/8036.

### DIFF
--- a/src/api/wix/WixToolset.Data/Symbols/SymbolDefinitions.cs
+++ b/src/api/wix/WixToolset.Data/Symbols/SymbolDefinitions.cs
@@ -322,6 +322,9 @@ namespace WixToolset.Data
                 case SymbolDefinitionType.FileSFPCatalog:
                     return SymbolDefinitions.FileSFPCatalog;
 
+                case SymbolDefinitionType.HarvestFiles:
+                    return SymbolDefinitions.HarvestFiles;
+
                 case SymbolDefinitionType.Icon:
                     return SymbolDefinitions.Icon;
 
--- a/src/test/msi/WixToolsetTest.MsiE2E/UIExtensionTests.cs
+++ b/src/test/msi/WixToolsetTest.MsiE2E/UIExtensionTests.cs
@@ -1,0 +1,26 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
+
+namespace WixToolsetTest.MsiE2E
+{
+    using Xunit;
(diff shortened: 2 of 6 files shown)


src/test/msi/WixToolsetTest.MsiE2E/UIExtensionTests.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    using Xunit;
+    using Xunit.Abstractions;
+
+    public class UIExtensionTests : MsiE2ETests
+    {
+        public UIExtensionTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper) { }
+
+        [Fact]
+        public void CanInstallAndUninstallFirewallRulesWithMinimalProperties()
```
robmen, line 13: The name of this test doesn't seem to apply to what it is testing. No?

src/test/msi/TestData/UIExtensionTests/LocalizedWixUI/product.wxs
```diff
@@ -0,0 +1,13 @@
+<!-- Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information. -->
```
robmen, line 1: Can we name this `LocalizedWixUI.wxs`?

I've stopped using `Product.wxs` and `Package.wxs` in my tests because when there are so many with the same name. The generic name makes it hard to figure out what value the .wxs file provides and which ones belong to which test (the directory name is _required_).

src/api/wix/WixToolset.Data/Symbols/SymbolDefinitions.cs
```diff
@@ -322,6 +322,9 @@ public static IntermediateSymbolDefinition ByType(SymbolDefinitionType type)
                 case SymbolDefinitionType.FileSFPCatalog:
                     return SymbolDefinitions.FileSFPCatalog;
 
+                case SymbolDefinitionType.HarvestFiles:
```
robmen, line 325: Should this change be in this PR? I'm fine if it tags along but it looks like it belongs with what I guess is your next fix. :)

robmen (changes requested):
The actual fix is laughably small.

src/api/wix/WixToolset.Data/Symbols/SymbolDefinitions.cs
barnson: Yes. It's been missing for a while, unfortunately, and shows up because I used `Files` in the test.

robmen (approved):

PR #510, opened 2024-03-18
Extensions can implement custom container compression

nirbar:
Closes wixtoolset/issues#8029
There's a working 7z implementation with this code in my WiX extension 
To build it follow these steps:
1. Clone https://github.com/nirbar/PanelSwWixExtension *with submodules*
2. Update src\global.json and src\Directory.Packages.props with WiX nupkg version
3. msbuild .\dirs.proj -restore -p:UnitTest=ContainerTemplateUT -p:EnableZipContainer=true


barnson:
I'll take a look, probably this weekend.

barnson:
This doesn't address the compressed blob we discussed earlier but the blocker is that requires the UX container to be a cab. That's a non-starter; I don't want the complexity of requiring two types of containers.

nirbar:
The container can be a blob, since the extension can manage all the required file metadata in it's BootstrapperExtensionData.xml node, or access BootstrapperApplicationData.xml.
About the ux.cab, how do you propose to compress the extension payloads and the manifests?

barnson:
WiX will have blobs baked in, so it needs a different solution.

PR #514, opened 2024-03-27
Allow default major upgrade to be rescheduled.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/8046.

### DIFF
--- a/src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
+++ b/src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
@@ -99,6 +99,7 @@ namespace WixToolset.Core.Link
                     SequenceTable = SequenceTable.InstallExecuteSequence,
                     Action = "RemoveExistingProducts",
                     After = "InstallValidate",
+                    Overridable = true,
                 },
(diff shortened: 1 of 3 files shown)


src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
```diff
@@ -126,7 +127,10 @@ private void AddSymbols(IntermediateSection section, params IntermediateSymbol[]
                 {
                     var symbolWithSection = new SymbolWithSection(section, symbol);
                     var fullName = symbolWithSection.GetFullName();
-                    this.Find.SymbolsByName.Add(fullName, symbolWithSection);
+                    if (!this.Find.SymbolsByName.ContainsKey(fullName))
```
robmen, line 130: Can't wait until we move to a C# version that supports the `.TryAdd()` method.

robmen (approved):

src/wix/WixToolset.Core/Link/AddDefaultSymbolsCommand.cs
barnson: Worse, it's `IDictionary`, which still doesn't have it...

PR #519, opened 2024-04-04
Report E_MOREDATA from Bal functions when string buffer too small

robmen:
WiX v4 (and probably v3) standardized on the error code E_MOREDATA when string buffers were too small instead of E_INSUFFICIIENT_BUFFER. This fixes v5 to match in a few missing cases.

Fixes wixtoolset/issues#8094


barnson (approved):
:shipit:

PR #521, opened 2024-04-27
CustomActionProxy: restore current directory after run.

rinrab:
The CustomActionProxy changes the current directory, but doesn't restore it after invocation.

It causes the problem that custom action cannot delete temp directory.

unnamed


robmen:
Thanks. Made a few style + error message tweaks in 546. This fix should get into WiX v5.0.1.

PR #525, opened 2024-05-21
Replace UpdateFileFacade loop with Parallel.ForEach

MBroholmA:
First: GitHub diff gets very confused around these changes. Visual Studio compare gives a much simpler view of the changes.

The main foreach facade loop in UpdateFileFacadesCommand.cs is replaced by a Parallel.ForEach loop. 
A new private class LocalData is used to store thread specific data and caches. Each thread gets a copy of the MsiAssemblyNameSymbols dictionary.
At the end of the loop the VariableCache's from each thread are merged back into the original.
Any exception breaks the loop and the first exception gets rethrown. 

The changes in BindDatabaseCommand.cs are to bring the CancellationToken forward to the Parallel.ForEach loop.

The current change doesn't limit the number of threads used. It could perhaps use the CabbingThreadCount parameter?

I found that a lot of the current tests cover this code, so I haven't written any new ones. I tried introducing an error on purpose which made a lot of them fail.

### DIFF
--- a/src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
+++ b/src/wix/WixToolset.Core.WindowsInstaller/Bind/BindDatabaseCommand.cs
@@ -6,6 +6,7 @@ namespace WixToolset.Core.WindowsInstaller.Bind
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Threading;
     using WixToolset.Data;
     using WixToolset.Data.Symbols;
     using WixToolset.Data.WindowsInstaller;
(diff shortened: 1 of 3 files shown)


MBroholmA:

barnson:
That diff is definitely ugly...I agree that there should be a way to limit the number of threads. I'd be fine reusing CabbingThreadCount. @robmen, do you have a preference?

MBroholmA:
Should I go ahead and use the CabbingThreadCount or should it be a separate setting?

barnson:
That's fine; I can't imagine needing separate control.

MBroholmA:
I've rebased and added the use of CabbingThreadCount. I had to fix a small bug in CalculateCabbingThreadCount to get it to work correctly

barnson (approved):

barnson:
Thanks!

PR #526, opened 2024-05-29
Fix build -includepath help message

Kuba314:
The includepath argument was displayed as `-include` in help message while only `-includepath` is accepted.

Closes [wixtoolset/issues#8527](https://github.com/wixtoolset/issues/issues/8527)


Kuba314:
recheck

barnson:
Thanks. We're blocked on builds atm but we'll take this as soon as we can.

PR #527, opened 2024-05-31
Add required VS components and configure VS in CI

campersau:
Fixes https://github.com/wixtoolset/issues/issues/8211 by implementing option 1. (Reverts https://github.com/wixtoolset/wix/pull/123)
Adds around 2-4 mins to build time.
Could be easily backported to v5 and v4.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,9 +32,9 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


robmen:
We're still debating whether this is the direction we want to go.

barnson:
We went with using the v143 toolset: https://github.com/wixtoolset/wix/pull/528

PR #528, opened 2024-06-10
Build native code with VC v143 toolset only.

barnson:
v143 output is compatible back to VS2017 (at least).

Fixes https://github.com/wixtoolset/issues/issues/8211.


robmen:
Should also update `README.md` to remove old references.

robmen (dismissed):
Update `README.md` and should be good to go.

PR #529, opened 2024-06-16
Error message for 'logon as service' is not currently localised.

bevanweiss:
Error message for failure to grant 'logon as service' rights is not currently localised.
Added translations in line with existing defined locales, Google translate used, but with some tweaking to try to keep 'log on as service' as a specific term.


bevanweiss:

barnson (approved):

barnson:
Thanks!

PR #530, opened 2024-06-19
[REVIEW please] CreateGroups additions

bevanweiss:
Just a placeholder for the current state of development on the ability to add Groups. https://github.com/wixtoolset/issues/issues/8577

Not sure if it works yet, splitting out the tables cost me a little more time than I anticipated, and I'm unsure that the decompile is right.

I'll build out a few more of the build test cases, to validate the expected preconditions etc (i.e. marking CreateGroup when it's not under a component, and similar).

Please raise any style / formatting issues, that way I've got the most time to address them :)

I've just noticed I've missed a couple of Wix4-.Wix6 references (around CustomAction names).  I'll tidy these up tomorrow.

Closes wixtoolset/issues#8577


src/wix/WixToolset.Converters/WixConverter.cs
```diff
@@ -224,6 +224,7 @@ public sealed class WixConverter
             { "WixCheckRebootRequired", "Wix4CheckRebootRequired_<PlatformSuffix>" },
             { "WixCloseApplications", "Wix4CloseApplications_<PlatformSuffix>" },
             { "WixRegisterRestartResources", "Wix4RegisterRestartResources_<PlatformSuffix>" },
+            { "ConfigureGroups", "Wix4ConfigureGroups_<PlatformSuffix>" },
```
bevanweiss, line 227: Looks like I was a bit over-eager.  I don't think a Wix6 feature needs a converter...

src/ext/Util/ca/scaexec.cpp
```diff
@@ -715,35 +860,127 @@ static HRESULT RemoveUserInternal(
...
+    LPWSTR pwzGroupDomain = NULL;
+    LPCWSTR wz = NULL;
+    PDOMAIN_CONTROLLER_INFOW pDomainControllerInfo = NULL;
+
+    //
+    // Remove the Group if the group was created by us.
+    //
+    if (!(SCAG_DONT_CREATE_GROUP & iAttributes))
+    {
+        if (wzDomain && *wzDomain)
```
bevanweiss, line 884: This entire DsGetDcName logic should be replaced by a GetServerName function call

src/ext/Util/ca/scaexec.cpp
```diff
@@ -715,35 +860,127 @@ static HRESULT RemoveUserInternal(
...
+            er = ::DsGetDcNameW(NULL, pwzDomain, NULL, NULL, DS_FORCE_REDISCOVERY, &pDomainControllerInfo);
         }
 
         if (ERROR_SUCCESS == er && pDomainControllerInfo->DomainControllerName)
         {
             // Skip the \\ prefix if present.
             if ('\\' == *pDomainControllerInfo->DomainControllerName && '\\' == *pDomainControllerInfo->DomainControllerName + 1)
             {
-                *ppwzServerName = pDomainControllerInfo->DomainControllerName + 2;
+                *ppwzServerName = _wcsdup(pDomainControllerInfo->DomainControllerName + 2);
```
bevanweiss, line 974: I suspect _wcsdup(...) is not the preferred function / macro to use.  But my quick google couldn't find a 'nicer' option.

src/ext/Util/ca/scaexec.cpp
```diff
@@ -1231,3 +1468,414 @@ extern "C" UINT __stdcall RemoveUser(
...
+        pGroupInfo1->grpi1_name = pwzName;
+        pGroupInfo1->grpi1_comment = NULL;
+
+        // Set the groups's comment
+        ApplyComment(iAttributes, pwzComment, &pGroupInfo1->grpi1_comment);
+
+        //
+        // Create the Group
+        //
+        GetServerName(pwzDomain, &pwzServerName);
```
bevanweiss, line 1549: I've forgotten to ReleaseStr(pwzServerName) in the LExit here...

src/ext/Util/ca/scagroup.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+    LPWSTR pwzData = NULL;
+
+    BOOL fGroupGroupExists = FALSE;
+
+    SCA_GROUP *psg = NULL;
+
+    INSTALLSTATE isInstalled, isAction;
+
+    if (S_OK != WcaTableExists(L"Wix4Group"))
```
bevanweiss, line 204: Check for Wix6Group needed here as well

src/ext/Util/ca/scagroup.cpp
```diff
@@ -0,0 +1,538 @@
...
+    {
+        WcaLog(LOGMSG_VERBOSE, "Wix4Group Table does not exist, exiting");
+        ExitFunction1(hr = S_FALSE);
+    }
+
+    if (S_OK == WcaTableExists(L"Wix4GroupGroup"))
```
bevanweiss, line 210: Should be Wix6GroupGroup

src/ext/Util/ca/scaexec.cpp
bevanweiss: StrAllocString(dest, src, 0 [size of src]) seems like the winner

bevanweiss:
@barnson Are there any additional integration tests that you think would be good to have beyond what I've already got in this?
There's a small set that I haven't yet put in that would be around the domain functionality (similar to the new user one I've added on the other PR #547), I think that would be:

1. Creating/removing a domain group
2. Adding/removing a domain user from a domain group
3. Adding/removing a domain group from a domain group
4. Adding/removing a domain user from a local group
5. Adding/removing a domain group from a local group

I know there's a heap of code in this PR, however there's not really many easily separable portions that would be meaningful on their own.  However, let me know if there's any way you can see that I could split it into more easily reviewable/pull-able portions.

barnson:
I'm not a domain expert but the tests look reasonable to me.

robmen (dismissed):

robmen:
One new test failed:

```log
[xUnit.net 00:04:46.85]     WixToolsetTest.MsiE2E.UtilExtensionGroupTests.FailsIfRestrictedDomain [FAIL]
  Failed WixToolsetTest.MsiE2E.UtilExtensionGroupTests.FailsIfRestrictedDomain [16 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at WixToolsetTest.MsiE2E.UtilExtensionGroupTests.FailsIfRestrictedDomain() in D:\a\wix\wix\src\test\msi\WixToolsetTest.MsiE2E\UtilExtensionGroupTests.cs:line 159
```

This is the test:

```cs
// Verify that a group cannot be created on a domain on which you don't have create user permission.
[RuntimeFact]
public void FailsIfRestrictedDomain()
{
    var productRestrictedDomain = this.CreatePackageInstaller("ProductRestrictedDomain");

    string logFile = productRestrictedDomain.InstallProduct(MSIExec.MSIExecReturnCode.ERROR_INSTALL_FAILURE, "TESTDOMAIN=DOESNOTEXIST");

    // Verify expected error message in the log file
    Assert.True(LogVerifier.MessageInLogFile(logFile, "CreateGroup:  Error 0x8007054b: failed to find Domain DOESNOTEXIST."));
}
```

bevanweiss:
Thanks, it looks like it's the log error message that couldn't be found.
I'll have a look into it.

Any chance there's a way to retrieve the log file from the Github Action?

robmen:
> Any chance there's a way to retrieve the log file from the Github Action?

Not today. E2E Burn logs are captured, but MSI E2E logs are not. 😢

bevanweiss:
> > Any chance there's a way to retrieve the log file from the Github Action?
> 
> Not today. E2E Burn logs are captured, but MSI E2E logs are not. 😢

No worries.
I did see what looked like that (Burn but not MSI) from the action.

And in your recent PR to update the runner actions it looked like the master repo was exempted from exporting those artefacts to prevent secret leakage also.

I just wanted to avoid a ‘works for me..’ situation.

robmen:
The `master` branch handling is for official builds and is (will be) done differently in WiX v6. You can ignore all references to `master` in the conditions.

bevanweiss:
Drats.  This looks like an issue with the revised Domain->Server method.
I think the intent of this method has changed, my understanding of the original version was that given a Domain, it would identify an applicable Domain Server (Domain Controller, DC) that should be used to direct the domain requests to.
If an invalid Domain was provided to the original method, it would return an error, indicating this domain was not known, and no server could be found.
The revised version of this method appears to just return the domain name as the server, and then because the HRESULT error from the NetAPI call (to get the domain server) is overwritten with the StringCopy result, the method itself returns all happy indications.

I think this GetDomainFromServerName doesn't make sense as-is... since we are explicitely given the domain.  It's the domain server that we need to identify (i.e. DC01.DOMAIN / PDC.DOMAIN etc).
And a side effect of this was the validation of the domain (i.e. if the domain doesn't exist, there can be no valid domain server for it, and hence the result value from the method should indicate this).
It was this PR that changed the method: https://github.com/wixtoolset/wix/pull/548

Unsure how I skipped this test on re-basing after that PR landed though...

bevanweiss:
> One new test failed:

Fixed.  I went back to the earlier GetDomainServerName() method, since this provides a usable HRESULT if the domain doesn't actually exist.

robmen (changes requested):

robmen (approved):

PR #535, opened 2024-06-30
Add a few more checks on rollback of Util User.

bevanweiss:
Add a few more checks on rollback of Util User.
~Tidy up the memory management around the ::NetUserGetInfo to align with MS guidance (https://learn.microsoft.com/en-us/windows/win32/api/lmaccess/nf-lmaccess-netusergetinfo#examples)~ 
Removed memory tidyup, it's now in the more recent PR #547

### DIFF
--- a/src/test/burn/WixTestTools/UserVerifier.cs
+++ b/src/test/burn/WixTestTools/UserVerifier.cs
@@ -43,13 +43,13 @@ namespace WixTestTools
         /// <param name="userName"></param>
         /// <param name="password"></param>
         /// <remarks>Has to be run as an Admin</remarks>
-        public static void CreateLocalUser(string userName, string password)
+        public static void CreateLocalUser(string userName, string password, string comment = "")
         {
             DeleteLocalUser(userName);
             UserPrincipal newUser = new UserPrincipal(new PrincipalContext(ContextType.Machine));
             newUser.SetPassword(password);
             newUser.Name = userName;
-            newUser.Description = String.Empty;
+            newUser.Description = comment;
             newUser.UserCannotChangePassword = true;
             newUser.PasswordNeverExpires = false;
             newUser.Save();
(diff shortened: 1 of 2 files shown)


bevanweiss:
I'm a little worried that the rollback tests for the Util User may not actually be functional.
When I uncomment the Assert's in ConfigureUsers, CreateUser and CreateUserRollback and run the rollback tests I'm only seeing the ConfigureUsers assert triggered.

My expectation was that all three would be triggered, the Configure would have setup the Create and the Rollback CAs, then the CreateUser should have triggered, Completed, then the CAFail would have triggered sometime later, which should have then resulted in the CreateUserRollback triggering.
Whilst the ConfigureUsers completes fine, neither of the other CAs appear to eventuate (in the ProductFail / ProductCommentFail tests).

barnson:
Rollback is scheduled only in this if block:

    if ((USER_EXISTS_YES == ueUserExists) || (USER_EXISTS_INDETERMINATE == ueUserExists) || !(psu->iAttributes & SCAU_DONT_CREATE_USER))

bevanweiss:
> Rollback is scheduled only in this if block:
> 
> ```
> if ((USER_EXISTS_YES == ueUserExists) || (USER_EXISTS_INDETERMINATE == ueUserExists) || !(psu->iAttributes & SCAU_DONT_CREATE_USER))
> ```

hmmm.... so in the Rollback Test (ProductFail / ProductCommentFail) I should not have expected CreateUser / CreateUserRollback to be executed?
I was really expecting to have CreateUser for each user (at least testName1 and testName2 which were to be created by the MSI).
And then post CAFail, I was expecting CreateUserRollback for each CreateUser that had actually modified state.

barnson:
I haven't looked at the tests yet. I just remembered that User is different than a lot because rollback of a removal isn't possible.

robmen (approved):

PR #536, opened 2024-06-30
A number of minor fixes for WiX v6

barnson:
* Remove unused const.
* For fields, "" != null - fixes https://github.com/wixtoolset/issues/issues/8558
* Exclude detect-only Upgrades from default-feature - fixes https://github.com/wixtoolset/issues/issues/8125

### DIFF
--- a/src/burn/engine/registration.cpp
+++ b/src/burn/engine/registration.cpp
@@ -5,7 +5,6 @@
 
 // constants
 
-const LPCWSTR REGISTRY_RUN_KEY = L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Run";
 const LPCWSTR REGISTRY_RUN_ONCE_KEY = L"SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\RunOnce";
 const LPCWSTR REGISTRY_BUNDLE_DISPLAY_ICON = L"DisplayIcon";
 const LPCWSTR REGISTRY_BUNDLE_DISPLAY_VERSION = L"DisplayVersion";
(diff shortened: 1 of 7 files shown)


barnson:
@robmen, please look (especially) at 845b9efe8db82746f9589738389226b282c8a07a.

src/wix/WixToolset.Core/Linker.cs
```diff
@@ -155,6 +148,17 @@ public Intermediate Link(ILinkContext context)
...
+                // If there are no authored features, assign the referenced components to
+                // the StdLib default feature, then re-run resolve to pick up the default
+                // feature from the StdLib.
+                {
+                    var command = new AssignDefaultFeatureCommand(find, resolve.ResolvedSections);
+                    command.Execute();
+                }
+
+                resolve = new ResolveReferencesCommand(this.Messaging, find.EntrySection, find.SymbolsByName);
+                resolve.Execute();
```
robmen, line 160: The other commits look fine but I want to look at a different solution than calling resolve references twice. This could introduce lots of complexity.

robmen (changes requested):
Might consider sending the other changes through and leave the fix for 845b9efe8db82746f9589738389226b282c8a07a for more consideration late.

src/wix/WixToolset.Core/Linker.cs
barnson: Please enjoy. Would you like an issue to track?

robmen: We should always open an issue to track since we'll forget about the issue otherwise. 😄

robmen (approved):

PR #542, opened 2024-07-12
Fix "wix extension list" to correctly display machine-wide extensions

robmen:
Fixes wixtoolset/issues#8625

### DIFF
--- a/src/wix/WixToolset.Core.ExtensionCache/ExtensionCacheManager.cs
+++ b/src/wix/WixToolset.Core.ExtensionCache/ExtensionCacheManager.cs
@@ -83,51 +83,54 @@ namespace WixToolset.Core.ExtensionCache
 
             (var extensionId, var extensionVersion) = ParseExtensionReference(extension);
 
-            var cacheFolder = this.GetCacheFolder(global);
+            var cacheFolders = this.GetCacheFolders(global);
 
-            var searchFolder = Path.Combine(cacheFolder, extensionId, extensionVersion);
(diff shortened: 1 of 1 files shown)


src/wix/WixToolset.Core.ExtensionCache/ExtensionCacheManager.cs
```diff
@@ -149,6 +152,18 @@ private string GetCacheFolder(bool global)
...
         }
 
+        private IEnumerable<string> GetCacheFolders(bool global)
+        {
+            if (this.cacheLocations == null)
+            {
+                this.cacheLocations = this.ExtensionManager.GetCacheLocations();
+            }
+
+            var cacheLocations = this.cacheLocations.Where(l => global || l.Scope == ExtensionCacheLocationScope.Project).OrderBy(l => l.Scope).Select(l => l.Path);
```
robmen, line 162: @barnson I've debated making the `.Where()` an `XOR` instead of `OR`. Right now (with the `OR`) the code shows both global _and_ project-level extensions if scanning globally. In other words, you see all available extensions.

The alternative is to say when looking global do NOT include project-level extensions (switch to an `XOR`). Thoughts?

barnson: `dotnet tool list` says `List tools installed globally or locally.` so inclusive is consistent with that.

barnson (approved):

PR #547, opened 2024-07-13
Util:user - Add domain runtime test skipping, and integration test including Domain="[valid domain]"

bevanweiss:
As the resolution of Domain User creation is a bigger fish than i'd like this week, i've dramatically scaled back this PR.
It now just introduces the Runtime conditional for domain membership for tests.
And adds a test (that if executed fails) showing the current deficiency in the User creation (when Domain is specified).
I'll use the Create Group changes to bring in altered treatment of Domain / Local Membership (in addition to the Domain / Local Group Creation / Removal).
Then it should be relatively easy to swing across the User handling in a subsequent PR.

Fixes:
- nothing... just includes test revealing issue with Domain User creation

### DIFF
--- a/src/test/burn/WixTestTools/RuntimeFactAttribute.cs
+++ b/src/test/burn/WixTestTools/RuntimeFactAttribute.cs
@@ -3,15 +3,20 @@
 namespace WixTestTools
 {
     using System;
+    using System.DirectoryServices.ActiveDirectory;
     using System.Security.Principal;
     using WixInternal.TestSupport.XunitExtensions;
 
     public class RuntimeFactAttribute : SkippableFactAttribute
     {
         const string RequiredEnvironmentVariableName = "RuntimeTestsEnabled";
+        const string RequiredDomainEnvironmentVariableName = "RuntimeDomainTestsEnabled";
 
         public static bool RuntimeTestsEnabled { get; }
         public static bool RunningAsAdministrator { get; }
+
+        public static bool RuntimeDomainTestsEnabled { get; }
+        public static bool RunningInDomain { get; }
 
         static RuntimeFactAttribute()
         {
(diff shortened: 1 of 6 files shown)


robmen:
@bevanweiss You've been doing some good work in PRs and definitely helping uncover some bugs. Now, I wanted to include the fixes to memory issues described in 8576 into v5.0.1, so I refactored some of your work into PR #548. That refactoring makes it easy to port. I didn't want you to think I was ignoring the work you've done here.

Also, some general feedback to help get some of your PRs across the finish line.

1. I would have preferred commits 1,2 and 6 combined and sent in a one-commit PR. That way the PR tackles a single issue with a minimal isolated change that can be reviewed and committed. That is essentially what I did in #548.
2. I left a comment in the 8629, but switching to `Impersonate="yes"` will break a lot. The existing CustomActions cannot be changed that way.

robmen (changes requested):
@bevanweiss You've been doing some good work in PRs and definitely helping uncover some bugs. Now, I wanted to include the fixes to memory issues described in 8576 into v5.0.1, so I refactored some of your work into PR https://github.com/wixtoolset/wix/pull/548. That refactoring makes it easy to port. I didn't want you to think I was ignoring the work you've done here.

Also, some general feedback to help get some of your PRs across the finish line.

1. I would have preferred commits 1,2 and 6 combined and sent in a one-commit PR. That way the PR tackles a single issue with a minimal isolated change that can be reviewed and committed. That is essentially what I did in https://github.com/wixtoolset/wix/pull/548.
2. I left a comment in the 8629, but switching to Impersonate="yes" will break a lot. The existing CustomActions cannot be changed that way.

bevanweiss:
> 1. I would have preferred commits 1,2 and 6 combined and sent in a one-commit PR. That way the PR tackles a single issue with a minimal isolated change that can be reviewed and committed. That is essentially what I did in [Fix faulty memory access in Util's User custom actions #548](https://github.com/wixtoolset/wix/pull/548).
> 2. I left a comment in the 8629, but switching to Impersonate="yes" will break a lot. The existing CustomActions cannot be changed that way.

Once your changes for the memory items are in, I'll look to squash (and maybe split) the commits I've got here.
If there's anything left of the memory items, I'll put them into a different PR (the split).
And then after further testing (and logs to back it up if it works as I think I'm seeing) I'll squash the other commits back into this PR.

src/test/burn/WixTestTools/RuntimeDomainFactAttribute.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
+
+namespace WixTestTools
+{
+    using System;
+    using System.DirectoryServices.ActiveDirectory;
+    using System.Security.Principal;
+    using WixInternal.TestSupport.XunitExtensions;
+
+    public class RuntimeDomainFactAttribute : RuntimeFactAttribute
```
robmen, line 10: Rather than create another `[Fact]` type, what about adding `RequiresDomain` bool property (or some better name) to the `RuntimeFact` attribute?

robmen (changes requested):
I know you're planning to rebase this change. I had an additional idea about not adding too many `[Fact]` types.

src/test/burn/WixTestTools/RuntimeDomainFactAttribute.cs
bevanweiss: done, because Named properties on attributes are only updated after the empty constructor has been called the code doesn't look very nice... but I think if using the same Attribute, having properly named attribute parameter(s) is better than just having nameless constructor parameters.

robmen:
@bevanweiss I'm uncertain what to do with the additional test given this statement in the commit message:

> This test currently fails under a Domain workstation.

If the test is failing, we should not have it dormant, waiting to surprise someone (future us, mostly likely) when the tests are finally run on a domain.

bevanweiss:
> @bevanweiss I'm uncertain what to do with the additional test given this statement in the commit message:
> 
> > This test currently fails under a Domain workstation.
> 
> If the test is failing, we should not have it dormant, waiting to surprise someone (future us, mostly likely) when the tests are finally run on a domain.

Once the Group Domain stuff is in, then I'll update the User stuff to align with it (i.e. separate Custom Actions for Domain vs Local actions).  In light of that, I'll mark this PR as 'Draft', so it's a placeholder to remind me to do this.

PR #548, opened 2024-07-15
Fix faulty memory access in Util's User custom actions

robmen:
Generally, clean up the handling of getting the domain from a server name by centralizing and simplifying it behind an improved GetDomainFromServerName() based on the buggy GetServerName().

Fixes wixtoolset/issues#8576

### DIFF
--- a/src/ext/Util/ca/utilca.cpp
+++ b/src/ext/Util/ca/utilca.cpp
@@ -1,3 +1,59 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
 
 #include "precomp.h"
+
+HRESULT GetDomainFromServerName(
+    __deref_out_z LPWSTR* ppwzDomainName,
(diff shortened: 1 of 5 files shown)


src/ext/Util/ca/utilca.cpp
```diff
@@ -1,3 +1,59 @@
...
+
+LExit:
+    // Note: we overwrite the error code here as failure to contact domain controller above is not a fatal error.
+    if (wz && *wz)
+    {
+        hr = StrAllocString(ppwzDomainName, wz, 0);
+    }
+    else // return NULL the server name ended up empty.
+    {
+        ReleaseNullStr(*ppwzDomainName);
```
barnson, line 49: Should we be freeing it too?

barnson (changes requested):
OK if freeing is OK

src/ext/Util/ca/utilca.cpp
barnson: It's internal, so it's fine; more of a general question about assuming DUtil memory management...

barnson (approved):

PR #549, opened 2024-07-15
MSMQ fixup

bevanweiss:
Fix up lack of WIX CUSTOM_ACTION_DECORATION wrappers
Add basic test for Msmq install/uninstall.
Add new RuntimeTest skipper for Windows Server / Optional Features.

Fixes wixtoolset/issues#8628

### DIFF
--- a/src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
+++ b/src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
@@ -1,0 +1,56 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
+
+namespace WixTestTools
+{
+    using System;
+    using System.Globalization;
+    using System.Linq;
+    using System.Management;
+    using System.Collections.Generic;
+    using System.Security.Principal;
+    using WixInternal.TestSupport.XunitExtensions;
+
+    public class RuntimePrereqFeatureFactAttribute : RuntimeFactAttribute
+    {
+        public static HashSet<string> OptionalFeatures = new(StringComparer.OrdinalIgnoreCase);
+        public static HashSet<string> ServerFeatures = new(StringComparer.OrdinalIgnoreCase);
+        static RuntimePrereqFeatureFactAttribute()
+        {
+            AddFeaturesToSet(ServerFeatures, "Win32_ServerFeature");
+            AddFeaturesToSet(OptionalFeatures, "Win32_OptionalFeature");
+        }
+
+        private static void AddFeaturesToSet(HashSet<string> featureSet, string featureSetName)
+        {
+            try
+            {
+                var objMC = new ManagementClass(featureSetName);
+                var objMOC = objMC?.GetInstances();
+                if (objMOC is not null)
+                {
+                    foreach (var objMO in objMOC)
+                    {
+                        string featureName = (string)objMO.Properties["Name"].Value;
+                        if ((uint)objMO.Properties["InstallState"].Value == 1)
+                        {
+                            featureSet.Add(featureName);
+                        }
+                    }
+                }
+            }
+            catch
+            {
+            }
+        }
+
+        public RuntimePrereqFeatureFactAttribute(params string[] prerequisiteFeatures) : base()
+        {
+            var missingRequirements = prerequisiteFeatures.Select(x => x).Where(x => !ServerFeatures.Contains(x) && !OptionalFeatures.Contains(x));
+
+            if (missingRequirements.Any())
+            {
+                this.Skip = "This test is missing the following Feature pre-requisites: " + String.Join(", ", missingRequirements);
+            }
+        }
+    }
+}
--- a/src/test/msi/WixToolsetTest.MsiE2E/MsmqExtensionTests.cs
+++ b/src/test/msi/WixToolsetTest.MsiE2E/MsmqExtensionTests.cs
@@ -1,0 +1,27 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information.
+
+namespace WixToolsetTest.MsiE2E
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using WixTestTools;
+    using Xunit.Abstractions;
+
+    public class MsmqExtensionTests : MsiE2ETests
+    {
+        public MsmqExtensionTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
+        {
+        }
+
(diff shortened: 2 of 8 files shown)


src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace WixTestTools
+{
+    using System;
+    using System.Linq;
+    using System.Management;
+    using System.Collections.Generic;
+    using System.Security.Principal;
+    using WixInternal.TestSupport.XunitExtensions;
+
+    public class RuntimePrereqFeatureFactAttribute : RuntimeFactAttribute
```
robmen, line 12: This is cool but I don't think the functionality to install pre-reqs should be done via the `[Fact]` attribute. It doesn't need to run per-Fact. Instead, it could run once as part of a fixture configuration and be used by all the tests. So, consider moving the functionality to WixInternal.TestSupport and we may need to add the "runtime test check" (env var check) there as well so it can be disabled the same way.

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
```diff
@@ -0,0 +1,65 @@
...
+                    {
+                        string featureName = (string)objMO.Properties["Name"].Value;
+                        if ((uint)objMO.Properties["InstallState"].Value == 1)
+                        {
+                            ServerFeatures.Add(featureName);
+                        }
+                    }
+                }
+            }
+            catch {}
```
robmen, line 34: Nit: Braces should go on their own line.

src/test/msi/TestData/MsmqExtensionTests/MsmqFailLog.txt
robmen, line 1: Is this file needed? If so, ideally it shouldn't be binary.

robmen (changes requested):

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
bevanweiss: This does not install.
It only tests for the presence of the feature in the 'installed' state (I was too lazy to add Status check also, the compile cycle on WiX is brutal).

I just wanted to be able to have Facts which could exist in repo, and would be assessable, but didn't report as Failures if the scenario wasn't right (i.e. in most situations.. like possibly runners, and even the Windows Sandbox in this situation).

I think conditional Windows Feature installation from a Test Fixture is a good idea, but agree it would need to be specifically checked for by an Environment Variable check.

I think 'the shape' of the parameters here (a list of strings) is tricky to group into the existing RuntimeFactAttribute (using 'params' to avoid needing to explicitly declare the 'new string[]').  So i'd rather this one be separate (although happy for a different name), even if the Domain test can be 'pulled into' the existing RuntimeFactAttribute.

src/test/msi/TestData/MsmqExtensionTests/MsmqFailLog.txt
bevanweiss: It shouldn't have been in the commit... I had removed it once before, not sure how it snuck its way back in.

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
robmen: Yeah, I didn't realize how ugly `string[]` array looked when used in as a property on an `Attribute`.  All the `static`s should tackle the initialization time.

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+    using System.Linq;
+    using System.Management;
+    using System.Collections.Generic;
+    using System.Security.Principal;
+    using WixInternal.TestSupport.XunitExtensions;
+
+    public class RuntimePrereqFeatureFactAttribute : RuntimeFactAttribute
+    {
+        public static HashSet<string> OptionalFeatures = new();
+        public static HashSet<string> ServerFeatures = new();
```
robmen, line 15: How about passing in `StringComparer.OrdinalIgnoreCase` to the hash sets so we don't have to worry about matching the case Windows chooses to use?

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+                            ServerFeatures.Add(featureName);
+                        }
+                    }
+                }
+            }
+            catch
+            {
+            }
+
+            try
```
robmen, line 38: This block looks to be a duplicate of the one above. Might be good to create a private static function to take in the management class name and hash set to populate. There is enough code here to reduce the duplication.

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+            catch
+            {
+            }
+        }
+
+        public RuntimePrereqFeatureFactAttribute(params string[] prerequisiteFeatures) : base()
+        {
+            var missingRequirements = prerequisiteFeatures.Select(x => x).Where(x => !ServerFeatures.Contains(x) && !OptionalFeatures.Contains(x));
+
+            if (missingRequirements.Count() > 0)
```
robmen, line 63: Nit: `.Any()` is a bit more descriptive than `.Count() > 0`

src/test/msi/WixToolsetTest.MsiE2E/MsmqExtensionTests.cs
```diff
@@ -0,0 +1,25 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using WixTestTools;
+    using Xunit.Abstractions;
+
+    public class MsmqExtensionTests : MsiE2ETests
+    {
+        public MsmqExtensionTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper) { }
```
robmen, line 15: Nit: braces on separate lines.

robmen (changes requested):

src/test/msi/WixToolsetTest.MsiE2E/MsmqExtensionTests.cs
bevanweiss: done

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
bevanweiss: done

src/test/burn/WixTestTools/RuntimePrereqFeatureFactAttribute.cs
bevanweiss: done

robmen (approved):

robmen:
Cool, cool. Thanks much.

PR #551, opened 2024-07-15
Fix egregiously bad comments that were frustrating

barnson:
Bad comments are seductively misleading...


robmen (approved):
Lots of whitespace changes too.

barnson:
VSCode FTW...

PR #553, opened 2024-07-17
Initialize WOW64 in TouchFile custom action

robmen:
Fixes wixtoolset/issues#8638

### DIFF
--- a/src/ext/Util/ca/TouchFile.cpp
+++ b/src/ext/Util/ca/TouchFile.cpp
@@ -23,6 +23,7 @@ static HRESULT SetExistingFileModifiedTime(
     )
 {
     HRESULT hr = S_OK;
+#ifndef _WIN64
     BOOL fReenableFileSystemRedirection = FALSE;
 
     if (f64Bit)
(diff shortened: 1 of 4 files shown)


robmen:
`WcaInitializeWow64()` was a bit new to me, so I'm not sure this is the best way to use it. It felt like a real hassle given the way this code is structured so I'm open to suggestions if I misunderstood how `WcaInitializeWow64()` and friends were intended to be used.

barnson:
I'm actually curious why this is needed. A 32-bit package can't touch the 64-bit file system so this would only be useful for touching a file installed by somebody else, no?

src/ext/Util/ca/TouchFile.cpp
```diff
@@ -23,6 +23,7 @@ static HRESULT SetExistingFileModifiedTime(
     )
 {
     HRESULT hr = S_OK;
+#ifndef _WIN64
```
barnson, line 26: ifndef/else bugs me but I realize it's just a pet peeve.

barnson (approved):

src/ext/Util/ca/TouchFile.cpp
barnson: (as opposed to ifdef/else)

robmen: Agreed but all of the other WiX CA code that deals with WOW64 redirection uses the same `#ifndef _WIN64` pattern. So does the rest of the code in this file. So,I thought it would be even more confusing to have this be the one place it was not `#ifndef`.

barnson: Not all of it; RemoveFolderEx does it my way. :) But then it means it switches from ifdef to ifndef when there's no else clause. So, 50/50.

PR #554, opened 2024-07-30
Net8, and a bunch of other dependency updates

bevanweiss:
This boosts up almost all of the WiX dependencies to their latest versions.
Main issue that arose was around System.Memory.dll binding with heat.exe and the updated MSBuild libraries.
Matching the upstream MSBuild bindingRedirections for System.Memory.dll appeared to resolve this.

I've run through the various Integration tests (Burn + MSI) after the build completed successfully, all worked as expected.

None of the non net6.0 targets have been touched, so net4xx (framework) support should remain as it was.
Have also not run any 'out of flow' testing over various environments like .NET 4.7.2 etc however, so a library conflict somewhere may cause an as yet unrevealed issue.  Happy to run additional tests if there's info available for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ This repository contains the WiX Toolset codebase.
 
 | Individual components |
 | :-------------------- |
-| .NET 6.0 Runtime (Long Term Support) |
+| .NET 8.0 Runtime (Long Term Support) |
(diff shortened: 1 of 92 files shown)


robmen:
Thanks, but we're not moving to .NET 8.0 in WiX v6.

PR #555, opened 2024-08-02
EnsureTable() should bring in relevant Standard Actions which rely on the table declared

bevanweiss:
The RemoveFoldersEx custom action relies upon pushing rows to the standard MSI table, and then expected the RemoveFiles standard MSI action to pick up the rows and do the removal actions. However without install files, we don't add in RemoveFiles scheduling.

This PR adds extra functionality after symbols have been resolved, where it will add in Standard Actions if they have not already been included elsewhere in the authoring.  The EnsureTable symbols are checked, and the table name used as a lookup to what Standard Actions it should directly bring in.

fixes wixtoolset/issues#8199
fixes wixtoolset/issues#8632 (in the as mentioned Standard Actions case.. not for generic custom actions)

### DIFF
--- a/src/api/wix/WixToolset.Data/WindowsInstaller/WindowsInstallerStandard.cs
+++ b/src/api/wix/WixToolset.Data/WindowsInstaller/WindowsInstallerStandard.cs
@@ -11,6 +11,7 @@ namespace WixToolset.Data.WindowsInstaller
     {
         private static readonly Dictionary<string, WixActionSymbol> standardActionsById;
         private static readonly HashSet<string> standardActionNames;
+        private static readonly Dictionary<string, WixActionSymbol[]> standardActionsByTableName = new Dictionary<string, WixActionSymbol[]>();
 
         /// <summary>
         /// References:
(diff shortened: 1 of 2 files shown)


bevanweiss:
@barnson is this direction more of what you were looking for? (obviously incomplete with table->standard action maps)

I'm still less a fan, some of them like the AppId have quite a few associated Standard Actions.  It feels 'safer' to me if each one has to be explicitly brought in by any Extension that requires them (with an EnsureStandardAction symbol).

barnson:
I'm still not sure what the advantage of extensions manually specifying standard actions would be.

bevanweiss:
> I'm still not sure what the advantage of extensions manually specifying standard actions would be.

My revised version (PR title / leader post not yet changed) just looks at EnsureTable symbols and maps them across to Standard Actions (based on Windows online doco).
Is there an easier 'definitive' reference for the table->action(s) mapping?
Did you want it going beyond EnsureTable symbols to all tables with references in the section?

My concern is some of the tables, like AppSearch etc etc bring in a number of Standard Actions which have various slightly weird sequencing sometimes.
So if an extension has EnsureTable("AppSearch"), is it then reasonable to bring in both InstallUISequence/AppSearch and  InstallExecuteSequence/AppSearch?
Or, does it make more sense that the Extension itself should know, and should specifically bring in one or the other?

barnson:
> My concern is some of the tables, like AppSearch etc etc bring in a number of Standard Actions which have various slightly weird sequencing sometimes.

Right---and that's why I _don't_ want extensions to be messing around at that level. MSI is usually the one enforcing the weird sequencing requirements, so I want the core WiX tools to be keeper of that knowledge as much as possible. Ideally, it wouldn't be possible to create an MSI that violates sequencing rules.

PR #557, opened 2024-08-03
Ensure naked files take Subdirectory into account.

barnson:
Naked files generated their ids using the attributes that are common to both naked and clothed files. But naked files also support Subdirectory to magic up a subdirectory in a specified directory (Directory) or the default INSTALLFOLDER. That subdirectory needs to factor in to the generated file id (which is then used as the component id too). Without it, generated ids for files with the same name but from different Subdirectory values would be duplicated. (Authored file ids must also continue to be supported.)

Naked files now generate different file and component ids. :(

Fixes https://github.com/wixtoolset/issues/issues/8674

### DIFF
--- a/src/wix/WixToolset.Core/Compiler.cs
+++ b/src/wix/WixToolset.Core/Compiler.cs
@@ -5348,7 +5348,7 @@ namespace WixToolset.Core
                 }
             }
 
-            if (null == id)
+            if (null == id && !isNakedFile)
             {
                 id = this.Core.CreateIdentifier("fil", directoryId, name);
             }
@@ -5610,8 +5610,6 @@ namespace WixToolset.Core
             string condition = null;
             string subdirectory = null;
 
-            var keyPath = this.ParseFileElementAttributes(node, "@WixTemporaryComponentId", directoryId, diskId: CompilerConstants.IntegerNotSet, sourcePath, out var _, componentGuid: "*", isNakedFile: true, out var fileSymbol, out var assemblySymbol);
-
             if (!this.Core.EncounteredError)
             {
                 // Naked files have additional attributes to handle common component attributes.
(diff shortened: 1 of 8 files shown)


barnson:
@robmen, you enabled automerge but didn't approve.

robmen:
Yes. I'm still mulling over this one.

src/wix/WixToolset.Core/Compiler.cs
```diff
@@ -5661,13 +5659,43 @@ private void ParseNakedFileElement(XElement node, ComplexReferenceParentType par
 
                 directoryId = this.HandleSubdirectory(sourceLineNumbers, node, directoryId, subdirectory, "Directory", "Subdirectory");
 
-                this.Core.AddSymbol(new ComponentSymbol(sourceLineNumbers, fileSymbol.Id)
+                var keyPath = this.ParseFileElementAttributes(node, "@WixTemporaryComponentId", directoryId, diskId: CompilerConstants.IntegerNotSet, sourcePath, out var _, componentGuid: "*", isNakedFile: true, out var fileSymbol, out var assemblySymbol);
```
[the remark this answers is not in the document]
robmen: I _feel_ like this change is the type of "minimal change" we'd make in a patch release to reduce churn. However, we're not in that mode so my thinking is that more complete refactoring will do us better in the long run.

Those are my _feelings_. Feel free to disagree as I don't have the specific change in mind... but I'd try refactoring it.

robmen (changes requested):

src/wix/WixToolset.Core/Compiler.cs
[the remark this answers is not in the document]
barnson: I needed the name parsed correctly. (I was also allowing for additional parsing but it turned out only name was necessary.) I was not offended by cloning a symbol. Would you rather I break out name parsing?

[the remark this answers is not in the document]
robmen: If all you need is the name, breaking out name parsing makes a ton of sense to me.

src/wix/WixToolset.Core/Compiler.cs
```diff
@@ -5661,13 +5685,23 @@ private void ParseNakedFileElement(XElement node, ComplexReferenceParentType par
...
                 directoryId = this.HandleSubdirectory(sourceLineNumbers, node, directoryId, subdirectory, "Directory", "Subdirectory");
 
-                this.Core.AddSymbol(new ComponentSymbol(sourceLineNumbers, fileSymbol.Id)
+                this.ParseFileNamingAttributes(node, sourcePath, directoryId, isNakedFile: true, out var id, out var name, out var shortName, out var source);
+
+                // Now that we have all the data we need to generate a good id, do
+                // so and create a file and component symbol with the right data.
+                id = id ?? this.Core.CreateIdentifier("nkf", directoryId, name, condition, win64.ToString());
+
+                var keyPath = this.ParseFileElementOtherAttributes(node, id.Id, directoryId, diskId: CompilerConstants.IntegerNotSet, id, name, shortName, source, out var _, componentGuid: "*", isNakedFile: true, fileSymbol: out var fileSymbol, assemblySymbol: out var assemblySymbol);
```
robmen, line 5694: The explicit parameter declaration looks a little funny in the out parameters that have the same names in the `out var`, like: `fileSymbol: out var fileSymbol` and `assemblySymbol: out var assemblySymbol` 

I'd remove the redundancy.

robmen (approved):
I like this much better. One nit.

src/wix/WixToolset.Core/Compiler.cs
barnson: Yeah, that's an error if you use a named argument earlier in the list. So all stay or all go. Vote?

robmen: Oh, I didn't realize you had to use it in all the following arguments. It adds value to the ones earlier, so it's worth it. Plus, maybe someone else will learn something if they try to "clean them up" in the future. :)

barnson: Yeah, it's not great but you gotta balance "naked `false`s" against "`out`s at the end."

PR #566, opened 2024-09-03
Fix crash with use of File/@PatchGroup.

barnson:


src/api/wix/WixToolset.Data/WindowsInstaller/WindowsInstallerStandard.cs
robmen, line 1: I don't think the changes to this file should be in this commit. Best not to touch `WixToolset.Data` in a commit that may need to be backported.

robmen (changes requested):

robmen (approved):

PR #567, opened 2024-09-13
Fallback to per-machine appdata folder when per-user appdata folder is not found.

cornerbowlsoftware:
This update is an attempt to move past the "Failed to find local per-user appdata directory." when not found.  Issue: [https://github.com/wixtoolset/issues/issues/8715](https://github.com/wixtoolset/issues/issues/8715)

### DIFF
--- a/src/burn/engine/cache.cpp
+++ b/src/burn/engine/cache.cpp
@@ -223,7 +223,7 @@ extern "C" HRESULT CacheInitialize(
     {
         LogErrorString(hr, "Failed to find local per-user appdata directory.  Falling back to per-machine appdata.");
 
-        // Fallback to per-machine adddata directory
+        // Fallback to per-machine appdata directory
         hr = ShelGetFolder(&sczAppData, CSIDL_COMMON_APPDATA);
         ExitOnFailure(hr, "Failed to find local %hs appdata directory.", "per-machine");
     }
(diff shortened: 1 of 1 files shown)


cornerbowlsoftware:

cornerbowlsoftware:
@barnson can you please look at this PR when you have time?

barnson (changes requested):
I don't think this is the right change. The unelevated engine will always fail when caching to a per-machine path so it's not really a fallback. Off the top of my head, the lack of a per-user cache path should fail as quickly as possible when a per-user cache is needed.

cornerbowlsoftware:
> I don't think this is the right change. The unelevated engine will always fail when caching to a per-machine path so it's not really a fallback. Off the top of my head, the lack of a per-user cache path should fail as quickly as possible when a per-user cache is needed.

@barnson In my case, I am only installing per-machine and the GMSA account that is remote installing does not have a per-user cache causing the engine to fail.  Can we leave the values uninitialized then when they are needed for users installing per-user, fail when the values are read?

barnson:
That's what I'm thinking. During planning, the engine knows if a per-user cache location is needed so if the path is NULL, throw an error. That prevents an access denied failure after elevating.

cornerbowlsoftware:
@barnson I can certainly help with the task, however it seems like you are the best person for it, and I suspect this is a two liner, one line to ignore the missing value in cache.cpp.CacheInitialize then another to throw the error in cache.cpp.GetRootPath but you are the expert. Please let me know if there is anything I can do to either implement it or help out.  Thanks for looking at this and all your work on this project. I am grateful I no longer have to use InstallShield.

barnson:
OK, then I'll mark this `up for grabs` so interested parties know it's available to work on.

PR #572, opened 2024-10-04
Add overloads to support create-only Wixouts.

barnson:
This prevents the .NET ZipArchive (and friends) from keeping the whole thing in memory, to support updating when we don't need to update the Wixout when building a binary Wixlib.

### DIFF
--- a/src/api/wix/WixToolset.Data/WixOutput.cs
+++ b/src/api/wix/WixToolset.Data/WixOutput.cs
@@ -25,7 +25,7 @@ namespace WixToolset.Data
         }
 
         /// <summary>
-        /// 
+        ///
         /// </summary>
         public Uri Uri { get; }
 
@@ -189,7 +189,10 @@ namespace WixToolset.Data
         /// <returns>Stream to the data of the file.</returns>
         public Stream CreateDataStream(string name)
         {
-            this.DeleteExistingEntry(name);
+            if (this.archive.Mode == ZipArchiveMode.Update)
+            {
+                this.DeleteExistingEntry(name);
+            }
 
             var entry = this.archive.CreateEntry(name);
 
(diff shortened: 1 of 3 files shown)


src/api/wix/WixToolset.Data/WixOutput.cs
```diff
@@ -73,6 +91,19 @@ public static WixOutput Create(Uri uri, Stream stream)
             return new WixOutput(uri, archive, stream);
         }
 
+        /// <summary>
+        /// Creates a new file structure that can only be written to.
+        /// </summary>
+        /// <param name="uri"></param>
+        /// <param name="stream">Stream to write the file structure to.</param>
+        /// <returns>Newly created <c>WixOutput</c>.</returns>
+        public static WixOutput CreateNew(Uri uri, Stream stream)
```
robmen, line 100: Since we're trying to be minimal on the API surface, let's merge this CreateNew with the CreateNew above.

src/api/wix/WixToolset.Data/WixOutput.cs
```diff
@@ -60,6 +60,24 @@ public static WixOutput Create(string path)
             return WixOutput.Create(uri, stream);
         }
 
+        /// <summary>
+        /// Creates a new file structure on disk that can only be written to.
+        /// </summary>
+        /// <param name="path">Path to write file structure to.</param>
+        /// <returns>Newly created <c>WixOutput</c>.</returns>
+        public static WixOutput CreateNew(string path)
```
robmen, line 68: For the sake of minimal change and given we want to improve this API, let's hide this for now and make it `internal`.

robmen (changes requested):

robmen (approved):

PR #576, opened 2024-11-14
Not generating short file names that might result in wrong file match

nirbar:
Fixes https://github.com/wixtoolset/issues/issues/8832


robmen:
This change will introduce validation failures in MSI packages. Thus, this is not the correct solution to generate short wildcard filenames.

PR #577, opened 2024-12-03
Fix Heat Harvesting for x64 Dlls

Nico-Ko:
Fixes: https://github.com/wixtoolset/issues/issues/8803

### DIFF
--- a/src/tools/heat/UtilFinalizeHarvesterMutator.cs
+++ b/src/tools/heat/UtilFinalizeHarvesterMutator.cs
@@ -856,7 +856,7 @@ namespace WixToolset.Harvesters
                                         processed = true;
                                     }
                                 }
-                                else if (5 == parts.Length && String.Equals("win32", parts[4], StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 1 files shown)


Nico-Ko:

src/tools/heat/UtilFinalizeHarvesterMutator.cs
```diff
@@ -856,7 +856,7 @@ private void MutateComponents()
                                         processed = true;
                                     }
                                 }
-                                else if (5 == parts.Length && String.Equals("win32", parts[4], StringComparison.OrdinalIgnoreCase))
+                                else if (5 == parts.Length && (String.Equals("win32", parts[4], StringComparison.OrdinalIgnoreCase) || String.Equals("win64", parts[4], StringComparison.OrdinalIgnoreCase)))
```
bevanweiss, line 859: Is 'arm64' also a potential option here?

Nico-Ko: Seems like it could be an option here. However, it looks as if support for arm64 would require more changes throughout the whole repository. As Heat appears to be deprecated anyway it is maybe sufficient to just add support for x64 platforms here.

robmen: Agreed, @Nico-Ko.

robmen (approved):

PR #582, opened 2024-12-29
Use PerUserProgramFilesFolder for default INSTALLFOLDER in a per-user package

robmen:
Also, adds a standard directory for `PerUserProgramFilesFolder`. 

Fixes wixtoolset/issues#8101

### DIFF
--- a/src/api/wix/WixToolset.Data/Symbols/WixPackageSymbol.cs
+++ b/src/api/wix/WixToolset.Data/Symbols/WixPackageSymbol.cs
@@ -18,6 +18,8 @@ namespace WixToolset.Data
                 new IntermediateFieldDefinition(nameof(WixPackageSymbolFields.Manufacturer), IntermediateFieldType.String),
                 new IntermediateFieldDefinition(nameof(WixPackageSymbolFields.Attributes), IntermediateFieldType.Number),
                 new IntermediateFieldDefinition(nameof(WixPackageSymbolFields.Codepage), IntermediateFieldType.String),
+                new IntermediateFieldDefinition(nameof(WixPackageSymbolFields.Scope), IntermediateFieldType.Number),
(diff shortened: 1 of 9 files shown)


src/api/wix/WixToolset.Data/Symbols/WixPackageSymbol.cs
```diff
@@ -44,12 +44,9 @@ public enum WixPackageAttributes
     {
         None = 0x0,
         PerMachine = 0x1,
-    }
-
-    public enum WixPackageUpgradeStrategy
-    {
-        None = 0x0,
-        MajorUpgrade = 0x1,
+        PerUser = 0x2,
```
barnson, line 47: Is this a general approach, to store bit flags in an `Attributes`?

barnson (approved):

src/api/wix/WixToolset.Data/Symbols/WixPackageSymbol.cs
robmen: Yeah, we do that in several symbols. The alternative is a lot of bools. The important part is that the attributes are combined correctly. Previously, the UpgradeStrategy was stomping over the other attributes. We just didn't notice because `per-machine` and `major ugprade` were both `0x1`. 🤬

robmen: I've thought about this more and decided that using `Attributes` for these two fields is too error-prone because two enum concepts are being squished into a single-bit field. So, I'll introduce dedicated fields for scope and upgrade strategy.

barnson: Hey, we're not MSI, needing to bit-twiddle to fit into the registry... :)

barnson (approved):

PR #587, opened 2025-01-03
Updated TestComponentNative.dll to be statically linked against MSVCRT140.dll

bevanweiss:
Updated TestComponentNative.dll to be statically linked against MSVCRT140.dll to avoid any issues with this dependency not existing (i.e. as was seen with Windows Sandbox environment).


robmen:
I missed this in a previous commit. We should build this file as part of the test project, not check in a binary.

PR #588, opened 2025-01-05
Migrate COM+ E2E tests from raw binaries to source code.

bevanweiss:


bevanweiss:
@robmen this PR resolves the issue with binaries previously committed for the COMPlus test cases

Me including the binary DLLs was based on the precendent from here (https://github.com/wixtoolset/wix/tree/main/src/test/msi/TestData/Templates)
image

Is the source for the CustomActionFail.exe available?

robmen:
I doubt we have the source code `CustomActionFail.exe` (it came from ancient unmaintained tests in wix3). We should eventually rewrite it as you did here. 👍🏻

robmen (approved):

robmen:
recheck

robmen (approved):

PR #589, opened 2025-01-14
Support WiX3-compantible auto-guids for registry key path

nirbar:
Fix [#8896](https://github.com/wixtoolset/issues/issues/8896)
By default, generated GUID will be compatible with WiX4+
Added an attribute to generate WiX3-compatible GUID: `Component/@WiX3CompatibleGuid=yes/no`

### DIFF
--- a/src/wix/test/WixToolsetTest.CoreIntegration/MsiQueryFixture.cs
+++ b/src/wix/test/WixToolsetTest.CoreIntegration/MsiQueryFixture.cs
@@ -2,6 +2,7 @@
 
 namespace WixToolsetTest.CoreIntegration
 {
+    using System;
     using System.IO;
     using System.Linq;
     using WixInternal.Core.TestPackage;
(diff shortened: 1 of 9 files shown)


src/wix/test/WixToolsetTest.CoreIntegration/MsiQueryFixture.cs
```diff
@@ -214,6 +215,37 @@ public void PopulatesAppSearchTablesFromRegistrySearch64()
...
+                });
+
+                result.AssertSuccess();
+
+                Assert.True(File.Exists(msiPath));
+                var results = Query.QueryDatabase(msiPath, new[] { "Component" });
+
+                Assert.Equal(2, results.Length);
+                Assert.Contains(results, component => component.Contains("WiX3CompatibleGuid") && component.Contains("8BAF5399-2FD2-50B6-ABDA-98FB3A5BB148", StringComparison.InvariantCultureIgnoreCase));
+                Assert.Contains(results, component => component.Contains("WiX4CompatibleGuid") && !component.Contains("8BAF5399-2FD2-50B6-ABDA-98FB3A5BB148", StringComparison.InvariantCultureIgnoreCase));
```
bevanweiss, line 245: I wonder if this might be 'better' as a strict equivalency test to the Wix4CompatibleGuid... in this manner the test would both verify that it's not the same as Wix3Guid, but also that if a future (potentially inadvertant) change makes things different from Wix4Guid it would be detected by this test.

This comment might be moot if there are already Wix4Guid tests elsewhere (I haven't looked specifically), and arguably having a dedicated set of guid tests is probably desirable elsewhere also.  But something for right now would be good..

nirbar: @bevanweiss I added a test to check compatibility with WiX v4.0.0 auto-genertated GUIDs for registry values under HKLM, HKU, HKCR, and for a file under ProgramFilesFolder

PR #591, opened 2025-01-17
Raise conversion warning for After and Before attributes

bevanweiss:
Raise conversion warning for After and Before attributes, at least on Sequence and SetProperty elements

### DIFF
--- a/src/wix/WixToolset.Converters/WixConverter.cs
+++ b/src/wix/WixToolset.Converters/WixConverter.cs
@@ -211,7 +211,7 @@ namespace WixToolset.Converters
             { "WixSchedHttpUrlReservationsInstall", "Wix4SchedHttpUrlReservationsInstall_<PlatformSuffix>" },
             { "ConfigureIIs", "Wix4ConfigureIIs_<PlatformSuffix>" },
             { "UninstallCertificates", "Wix4UninstallCertificates_<PlatformSuffix>" },
-            { "InstallCertificates", "Wix4_<PlatformSuffix>" },
+            { "InstallCertificates", "Wix4InstallCertificates_<PlatformSuffix>" },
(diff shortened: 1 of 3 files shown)


robmen:
This fixes issue 8898, correct? If so, please always add the following line as the last part of the PR message. It ties the PR to the issue and auto-closes it so we don't have to remember to do so later. :)

```

Fixes wixtoolset/issues#ISSUE_NUMBER_HERE
```

bevanweiss:
> This fixes issue 8898, correct? If so, please always add the following line as the last part of the PR message. It ties the PR to the issue and auto-closes it so we don't have to remember to do so later. :)
> 
> ```
> 
> Fixes wixtoolset/issues#ISSUE_NUMBER_HERE
> ```

not quite...  from the issue

> Expected Result
The value WixQueryOsDirs should have been replaced with Wix4QueryOsDirs_$(sys.BUILDARCHSHORT).

This doesn't appear to be how the converter has previously handled these wix custom actions, and I've kept it similar.
So my PR just raises the warnings for the conversion locations, it doesn't automatically fix them.
Although that should be pretty straight forward to implement if that's the direction we'd want.

These would need to change to have `_$(sys.BUILDARCHSHORT)` in place of `_<PlatformSuffix>` for these
https://github.com/wixtoolset/wix/blob/523c66a62a619e6aa9f30070173ea33edfb5e328/src/wix/WixToolset.Converters/WixConverter.cs#L203
and an update of the attribute values, plus possibly removal of the warning.

robmen:
Yeah, I think we're still debating if `_$(sys.BUILDARCHSHORT)` is our final solution or if we're going to try to find a better way to tackle architecture specific custom actions. Thanks for picking up the Before/After attributes.

robmen (approved):

PR #594, opened 2025-01-22
Net8

bevanweiss:
Brings dependencies up to .NET 8 (whilst keeping backward support for earlier releases under .NET Standard 2.0).

fixes wixtoolset/issues#8909


barnson:
8909 is for WiX v7.

bevanweiss:
> 8909 is for WiX v7.

Understood.  It's easy enough to rebase and update patch versions of the .NET 8 references in this PR though.
And I already had it laying around.

PR #596, opened 2025-01-23
Add DirBrowseFlags to IisWebDirProperties

bevanweiss:
Adds in support for configuring IIS DirBrowseFlags to Web Directories.

Fixes wixtoolset/issues#170
Fixes wixtoolset/issues#2637
Fixes wixtoolset/issues#2917


bevanweiss:
I think it might be better to rename the Db column as a more generic 'Attributes' field, since this provides some additional bits for flags for future enhancement with other boolean settings (would have been useful if this was already in so I could have used some for these browse flags).

I think there is a test case for a 'bug' in here also (commented out).
<iis:WebDirProperties Id="TestCacheControlMaxAgeNull" **_CacheControlMaxAge="2147483648_"** />

will crash out database validation since the 2147483648 value == 0x80000000 the MSI_NULL_INTEGER...
Not sure on a good fix for it however.  
1. Perhaps just adding +1 / -1 in the compiler to avoid this specific value.  or 
2. A compiler error that specifically disallows this value without the confusing error around DB validation.

Since this column addition will break the merge module back compatibility (as I understand), it should probably be coupled with some other changes to IIS schema also (ScriptTimeout I think was the other)

bevanweiss:
This PR doesn't have the changes indicated around the Script timeouts.
I'll possibly look at this 'sometime soon'... but it's unlikely to be affecting too many WiX users.. so I might look at some of the accessibility items instead.

PR #598, opened 2025-01-28
Implement UseArtifactsOutput to mirror the .NET SDK

brantburnett:
Resolves https://github.com/wixtoolset/issues/issues/7401


brantburnett:
@robmen I do have a couple of particular questions about this PR when you have a moment.

1. Is the testing sufficient? What additional coverage would you like, more OutputTypes, more variations, or something included in WixE2E?
2. I couldn't find documentation on current error codes, so I picked WIX9000 and WIX9001 for these new MSBuild-related errors as a placeholder.

PR #604, opened 2025-02-15
Re-integrate XSDs

robmen:


barnson:
Shouldn't foo.wixext xsd live under ext\foo?

bevanweiss:
> Shouldn't foo.wixext xsd live under ext\foo?

I think this would then possibly require some kind of linkage (maybe xsd:include would work) from a fixed path to the 'unknown' variable number of extension xsd paths.
It does seem nicer to have the extension xsd files located next to their code though.

robmen:
I actually tried spreading them all out, but it was harder to keep them consistent. So, I thought putting them together would help.

barnson:
`harder to keep them consistent`

❓

robmen:
The XSDs are a bit of a mess right now in that some redefine basic types like `YesNo` and some use types from `wix.xsd`. When editing right now, one needs to search across the XSDs more often than not to keep them appropriately in sync. There is absolutely a future task here to improve the structure of the XSDs. I have a bunch of ideas, but I don't know if any are better than what we have right now.

barnson (approved):

PR #605, opened 2025-02-17
Convert xUnit tests to MSTest and increase use of traversal projects to increase build perf

barnson:
Work for https://github.com/wixtoolset/issues/issues/8685.


bevanweiss:
Is the migration from xUnit to MSTest due to extra functionality in MSTest, or an issue with xUnit?

xUnit does have support for the Microsoft Testing Platform
https://xunit.net/docs/getting-started/v3/microsoft-testing-platform

If it's just about parallel testing, it seems this could just be a case of missing parallel enabling in xUnit?
https://xunit.net/docs/configuration-files

I'm keen on having a try of this and potentially seeing how it goes with the code coverage extension.

barnson:
Getting .exe test assemblies was a necessary part of a multiproc build. Now that xUnit supports it, I wouldn't be opposed to using it but I found MSTest to offer a better developer experience (e.g., with an MSBuild SDK).

barnson:
@robmen The action runner VM got caught up in the long-running tests (Sql and Firewall). That's a downside of the segment scope though in this case, there's not much after `ext` to buffer the long-running builds. So even with 4 cores, we're not seeing big wins. We should still get some with traversal projects building everything.

robmen (approved):

PR #606, opened 2025-02-22
Append _wixcert_ to CA test for rollback check

bevanweiss:
Taken from wixtoolset/wix3/pull/546, but have left out the explicit length calculation from the pwzUniquePrefix string, and using this in the compare.

Fixes wixtoolset/issues#3976

### DIFF
--- a/src/ext/Iis/ca/scacert.cpp
+++ b/src/ext/Iis/ca/scacert.cpp
@@ -701,6 +701,10 @@ static HRESULT FindExistingCertificate(
     PCCERT_CONTEXT pCertContext = NULL;
     BYTE* pbCertificate = NULL;
     DWORD cbCertificate = 0;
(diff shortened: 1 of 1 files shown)


bevanweiss:
I think maybe the whole `_wixcert_x` portion should disappear (and so this PR isn't so valid).
I believe the thumbprint of the certificate should be sufficient.
Any objections to changing to Thumbprints to know what activities we should / have already done (for the install / rollback)?

robmen:
It's been a long time since I looked at the cert custom actions, but **IIRC** the thumbprint could be provided as a property so it wasn't necessarily a stable identifier that could be used to uninstall the certificate later. This is fuzzy, but it may be that the cert install behavior needs more of a re-design to address the suffix... which is why that issue was just hanging out unfixed.

PR #612, opened 2025-03-03
Fix internal UI bootstrapper application taskbar icon disappear

leo4707:
As https://github.com/wixtoolset/issues/issues/8135 metion, WixInternalUIBootstrapperApplication cannot show icon on taskbar correctly. Due to the usage of CreateWindowExW in WixToolset.BootstrapperApplications CreateMainWindow, WS_EX_TOOLWINDOW will make the icon disappear in taskbar refer to [document](https://learn.microsoft.com/en-us/windows/win32/winmsg/extended-window-styles). I consider that it is the reason why it cannot show properly.

Fixes https://github.com/wixtoolset/issues/issues/8135

### DIFF
--- a/src/ext/Bal/stdbas/WixInternalUIBootstrapperApplication.cpp
+++ b/src/ext/Bal/stdbas/WixInternalUIBootstrapperApplication.cpp
@@ -578,7 +578,7 @@ private:
             dwWindowStyle |= WS_VISIBLE;
         }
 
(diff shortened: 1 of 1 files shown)


leo4707:

leo4707:
recheck

bevanweiss:
If this would fix the issue, it'd be nice to add that information to the commit (it just means the issue will be automatically closed when the PR is merged).  I'm pretty sure you can also add it to the PR description for it to do the same (which prevents having to re-write commit history... not that branch commit history is especially sacred).

robmen (approved):

PR #613, opened 2025-03-04
Only referenced components get default feature.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/8882 Fixes https://github.com/wixtoolset/issues/issues/8939

### DIFF
--- a/src/api/wix/WixToolset.Data/WixStandardLibrary.cs
+++ b/src/api/wix/WixToolset.Data/WixStandardLibrary.cs
@@ -93,20 +93,6 @@ namespace WixToolset.Data
                 yield return section;
             }
 
-            // Default feature.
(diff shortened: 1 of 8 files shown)


robmen (approved):

barnson:
Yep, the re-resolve was only to resolve the generated references to the feature symbol in StdLib. Creating the symbol "manually" avoids that. Brute force ftw!

robmen:
I've convinced myself that this path is the right one for both "Default Feature" and "INSTALLFOLDER" because those two symbols serve as the central pillars of the MSI. They are fundamental enough that they can't just "exist in the StdLib", and that's okay.

Nice work. Will be great to have this bug(s) resolved.

PR #614, opened 2025-03-10
Set keypath when parsing naked File child elements

barnson:
Fixes https://github.com/wixtoolset/issues/issues/8976


barnson:
Trying for better title...

robmen (approved):

PR #617, opened 2025-03-18
Files harvesting: non-existent directory causes disruption of the harvesting process

olejniczak:
Fixes https://github.com/wixtoolset/issues/issues/8989


bevanweiss:
> Fixes #8989

Because WiX has the issues in a separate repo (to migrate from v3 onwards) you need to provide more context for Github to link them..
i.e.
Fixes wixtoolset/Issues#8989

PR #620, opened 2025-03-20
Introduce Open Source Maintenance Fee EULA

robmen:
Add EULA to .nupkgs for wixtoolset/issues#8974


src/setup/ThmViewerPackage/ThmViewerPackage.wxs
```diff
@@ -1,6 +1,6 @@
 ﻿<!-- Copyright (c) .NET Foundation and contributors. All rights reserved. Licensed under the Microsoft Reciprocal License. See LICENSE.TXT file in the project root for full license information. -->
 <Wix xmlns="http://wixtoolset.org/schemas/v4/wxs">
-  <Package Id="WixToolset.ThemeViewer.Package" Name="WiX Toolset Theme Viewer" Manufacturer="WiX Toolset" Language="1033" Version="!(bind.fileVersion.ThmViewerFile)">
+  <Package Id="WixToolset.ThemeViewer.Package" Name="WiX Toolset Theme Viewer" Manufacturer="$(Company)" Language="1033" Version="!(bind.fileVersion.ThmViewerFile)">
```
barnson, line 3: (and elsewhere): Should `$(Company)` be in the `Id`? I'm trying to decide if it's a positive or negative that different forks could upgrade each other...

barnson (changes requested):
I'm OK with either answer on my comment but didn't want the automerge to kick in...

src/setup/ThmViewerPackage/ThmViewerPackage.wxs
robmen: It can't be `$(Company)` specifically because `$(Company)` has spaces in it. However, I think this is the right value. If someone wants to truly install SxS they'll have to update more than just the `Package.Id` so I think this hints that there is more work to be done in that case.

barnson (approved):

PR #632, opened 2025-04-06
Support launching rundll32.exe as a safe executable if the dll it loads is in a secure location

nirbar:
Closes https://github.com/wixtoolset/issues/issues/9025

### DIFF
--- a/src/burn/engine/approvedexe.cpp
+++ b/src/burn/engine/approvedexe.cpp
@@ -3,6 +3,13 @@
 #include "precomp.h"
 
 
+// internal function declarations
+
+static HRESULT IsRunDll32(
+    __in BURN_VARIABLES* pVariables,
+    __in LPCWSTR wzExecutablePath
+    );
+
 // function definitions
 
 extern "C" HRESULT ApprovedExesParseFromXml(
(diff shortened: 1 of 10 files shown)


bevanweiss:
Does the existing `ApprovedExesVerifySecureLocation` already 'resolve' paths before comparing them to 'secure' locations?
If not, it might be worth considering this as part of the change (given the increased security surface area this PR imparts).

My thought is if the 'secure location' is `C:\Windows\System32\..\..\RealEvil\RunMyEvil.exe`, is it going to compare it to `C:\Windows\System32` and then say it's all secure and good to run... but when resolved this actually points to `C:\RealEvil\RunMyEvil.exe`...

I did try to find something like [PathResolve](https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-pathresolve), but couldn't see anything in the C source code.. only in C#

nirbar:
`ApprovedExesVerifySecureLocation` resolves paths.
I added a test for the path `C:\Program Files\..\a.exe`

robmen:
This is incredibly dumb by Nvidia, but thanks for helping support it. 😝

PR #638, opened 2025-05-12
Fix Width of lines to be equal to dialog width

MikeWilliams-UK:
This should fix https://github.com/wixtoolset/issues/issues/9058

DEBUG: Error 2826:  Control BottomLine on dialog PrepareDlg extends beyond the boundaries of the dialog to the right by 5 pixels

Width of all dialogs is 370
The length of the lines changed here was 373 which exceeds 370

I suspect that the reason I sometimes see "to the right by 5 pixels" or "to the right by 4 pixels" depends on the actual scaling applied.


MikeWilliams-UK:

MikeWilliams-UK:
Abandoning this pull request.

PR #641, opened 2025-05-26
Add exists option to ProductSearch Result for matching other search types

metsma:
(e.g. FileSearch, DirectorySearch, RegistrySearch)

Signed-off-by: Raul Metsma <raul@metsma.ee>


metsma:

src/burn/engine/search.cpp
```diff
@@ -1237,6 +1243,7 @@ static HRESULT MsiProductSearch(
     case BURN_MSI_PRODUCT_SEARCH_TYPE_LANGUAGE:
         type = BURN_VARIANT_TYPE_STRING;
         break;
+    case BURN_MSI_PRODUCT_SEARCH_TYPE_EXISTS: __fallthrough;
```
robmen, line 1246: Rather than fall through here and check the type again below (line 1255), we should just set the value appropriately in this `case`.

src/burn/engine/search.cpp
```diff
@@ -1245,6 +1252,11 @@ static HRESULT MsiProductSearch(
     hr = BVariantChangeType(&value, type);
     ExitOnFailure(hr, "Failed to change value type.");
 
+    if (BURN_MSI_PRODUCT_SEARCH_TYPE_EXISTS == pSearch->MsiProductSearch.Type)
+    {
+        value.llValue = value.llValue == INSTALLSTATE_ABSENT ? 0 : 1;
```
robmen, line 1257: I'd like to see parens around the ternary condition:

```
value.llValue = (value.llValue == INSTALLSTATE_ABSENT) ? 0 : 1;
```

And moved up to the correct `case`. :)

robmen (changes requested):
This is a pretty good code change. Thanks.

I'd like to see a test added as well to make sure we don't regress this behavior.

src/burn/engine/search.cpp
metsma: Isn't the value before BVariantChangeType string type?

src/burn/engine/search.cpp
metsma: The value is in this stage string format?

metsma: Do you want me to anything with this commit?
I am not using existing case because at this point the value is string ang later converted to int.

    hr = BVariantChangeType(&value, type);

and then int value is compared

    value.llValue = (value.llValue == INSTALLSTATE_ABSENT) ? 0 : 1;

PR #643, opened 2025-06-02
Update ErrorMessages.cs to refine duplicate identifiers message

cz-dev-ge:
Adds information about necessary neutral language definition.

If multiple language files are defined in burn, but no neutral language is given, this error message is returned. But without the proposed change it is not clear why this error appears.

See [Discussion 9081](https://github.com/orgs/wixtoolset/discussions/9081), too.


cz-dev-ge:
>
> I have read the CLA Document and I hereby sign the CLA
>

barnson (approved):

PR #644, opened 2025-06-05
Get tests listed as they're run.

barnson:


barnson:
Good bit already fixed in c7b981244cfe0dbca2284690bd5091aeda410de0.

PR #649, opened 2025-07-10
Fix for issue9118 - wrong elevation detection for non-admin accounts with UAC disabled

svg2003:
Fix for issue: https://github.com/wixtoolset/issues/issues/9118
When user is not in admin group and UAC is disabled, wix may wrongly detect process as elevated


svg2003:
PS. I also created a fix for 4.0.7., but seems you're not using branches, so not sure how to create PR for that. It shows me all missed commits
https://github.com/svg2003/wix/tree/issue9118-v407

robmen:
This is a problem, but this is not the correct fix for it. I've taken the issue in WiX v7 and should have a correct fix.

PR #653, opened 2025-07-20
Embiggen build perf.

barnson:
- Move ext tests to targets for better MSBuild scheduling.
- Parallelize long-running tests.
- Disable Firewall ICEs for dev builds; they're redundant with round-tripping and ICEs are mutex'd to make them painfully serial.


barnson:
Shaves >90 seconds off a full build.

PR #654, opened 2025-07-27
Fix for issue 9104 - Relation patch bundle check should not be done during uninstall

JasonLow8:
Fix for issue: https://github.com/wixtoolset/issues/issues/9104

This PR fixes the issue when the parent bundle is uninstalled, all related bundle is not uninstalled and orphaned. The request state is incorrect because it did not check for uninstall action.

### DIFF
--- a/src/burn/engine/plan.cpp
+++ b/src/burn/engine/plan.cpp
@@ -376,20 +376,6 @@ extern "C" HRESULT PlanDefaultPackageRequestState(
             break;
         }
     }
-    else if (BOOTSTRAPPER_RELATION_PATCH == relationType && BURN_PACKAGE_TYPE_MSP == packageType)
-    {
-        // For patch related bundles, only install a patch if currently absent during install, modify, or repair.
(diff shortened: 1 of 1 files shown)


JasonLow8:
Added failing test, which succeed upon code modification. 

But packageEv101 is still installed after bundleJ_patch is uninstalled, which is another bug on its own. A parent bundle is being uninstalled from the control panel, it will uninstall all bundle patch with "-burn.related.patch" parameter, before proceeding to uninstall the parent bundle. If an updated MSI is being supplied in a bundle patch, it will be left orphaned.

robmen:
I appreciate you opening issues tracking the additional problems you found while trying to fix the issue you are interested in, however, we can't go forward without a test that verifies the change you are making. The skipped test is fine if it shows the problem described in the new issue but we need a passing test to know that your current change is working correctly.

JasonLow8:
Added a test to verify the changes i have made. also left the test for the new issue i found.

JasonLow8:
Hi,

Is there anything left to do for this pull request to be reviewed?

PR #702, opened 2026-03-26
Also find MSBuild v18 (vs 2026) when running tests

martinh2011:
VswhereFindArguments in MsBuildrunner.cs only includes versions up to 17, excluding 18 (VS 2026). This PR changes that to include 18, too

fixes wixtoolset/issues#9269

PR #706, opened 2026-06-04
Target VS2026 and .NET 10 for WiX v8

barnson:
Closes #702


barnson:
@robmen Switching WixE2E to MSTest resolved the 15-minute delays on those MSBuild invocations. Ready to go now.

robmen (approved):

PR #707, opened 2026-06-09
Fix build on machines with only VS2026

avjts:
I couldn't build WiX on a machine with VS2026 (with 2022 toolset), because test runner searches for VS2022.


avjts:
Now I spotted that this is already fixed in #706

PR #711, opened 2026-07-06
UAC shield for per-machine configurable bundles.

### LINKED ISSUE #9308: WixStdBA should show UAC shield when configurable bundle scope is per-machine
### Task

`BAL_INFO_BUNDLE` doesn't know about scope, just `fPerMachine`, which is what WixStdBA uses to show the UAC shield. Make it smarter.

barnson:
Fixes https://github.com/wixtoolset/issues/issues/9308


dallmair:
@barnson I guess this is related to, but does not fix, wixtoolset/issues#9321 ?

robmen (approved):

PR #715, opened 2026-08-02
Add CancelWhenDeferred, like FailWhenDeferred.

barnson:
Useful for integration testing packages.


src/ext/Util/test/WixToolsetTest.Util/UtilExtensionFixture.cs
```diff
@@ -100,6 +100,21 @@ public void CanBuildCloseApplication()
...
+        public void CanBuildWithFailures()
+        {
+            var folder = TestData.Get(@"TestData\Failures");
+            var build = new Builder(folder, typeof(UtilExtensionFactory), new[] { folder });
+
+            var results = build.BuildAndQuery(BuildX64, "Binary", "CustomAction");
+            WixAssert.CompareLineByLine(
+            [
+                "Binary:Wix4UtilCA_X64\t[Binary data]",
+                "CustomAction:Wix4CancelWhenDeferred_X64\t1025\tWix4UtilCA_X64\tWixCancelWhenDeferred\t",
```
robmen: `Wix4CancelWhenDeferred`? Not `Wix7`?

barnson: Argh. I hate whoever came up with this wacky idea...and all because of stupid merge modules. Well, it would be `Wix8` right? But we still use `Wix4` for the Binary id. Maybe we should be rev'ing _that_ every release? Or do we need to change it if we add something? But not change something? 🤕

barnson: Oh wait, the Binary id gets a modularization guid, so it's "safe."

barnson: It's just the table names that need to be unique.

barnson: We have doc that uses `BinaryRef="Wix4UtilCA_$(sys.BUILDARCHSHORT)"` so we shouldn't change it unnecessarily. If I had a time machine, I'd have done this differently.

PR #719, opened 2026-09-21
Include Binary stream differences in msi transform.

### LINKED ISSUE #9342: wix msi transform does not include binary difference
### WiX Version

7.0.0

### .NET or MSBuild or Visual Studio Version

Visual Studio 18.10.0

### HeatWave Version

_No response_

### Repro Steps

1. Create a msi installer with a background image resource i.e. Resources/background.bmp
2. Create a second msi installer with a different image resource as 1 but same naming.
3. Execute `wix.exe msi transform -p main.msi main.otherbrand.msi -out sometransform.mst`

someTransform.mst should contain the image difference.

### Repro Repo

_No response_

### Actual Result

WIX V3 torch.exe was capabable to create a binary difference but the new
WIX V7 wix.exe approach does not.

see details:
https://github.com/orgs/wixtoolset/discussions/9335

### Expected Result

wix.exe msi transform main.msi main.otherbrand.msi -out sometransform.mst

should return a mst file with binary difference.

### Acknowledgements

- [x] I acknowledge that this is a fully completed bug report. It is not a question or attempt to get help debugging my issue (because those should be [sent to Discussions](https://github.com/orgs/wixtoolset/discussions/new?category=questions)).
- [x] I (or an organization I'm a member of) pay the [Open Source Maintenance Fee](https://opensourcemaintenancefee.org/) by [sponsoring the `wixtoolset` project](https://github.com/sponsors/wixtoolset) because I generate revenue from my use of the WiX Toolset.

tallha-khan:
## Summary
- Fixes `wix msi transform` omitting Binary table stream changes when two MSIs share the same Binary name but different contents (for example a rebranded background.bmp).
- Exports each MSI's object streams into separate folders so the binder compares real files instead of `FILE NOT EXPORTED`, including when `-x` is omitted.
- Adds a CoreIntegration regression test that builds two packages, creates an MST, applies it, and asserts the Binary stream is updated.

Fixes https://github.com/wixtoolset/issues/issues/9342

## Test plan
- [x] CanIncludeBinaryStreamDifferenceInTransform passed locally
- [x] Existing TransformFixture tests passed locally
- [x] Reproduced WIX0227 with WiX 7.0.0 on two MSIs that only differ by Binary content; local build created an MST that updates the stream from 8 bytes to 10 bytes


tallha-khan:

